PR #38, opened 2023-12-04
update xcode environment schema

0x401080:
Rationale: 
if I'm using environment variables from the [official tutorial](https://docs.vulkan.org/tutorial/latest/02_Development_environment.html#_setting_up_xcode) in Xcode, the result is below. 

Screenshot 2023-12-04 at 22 23 12

```
required extensions:
1
	VK_KHR_portability_enumeration
available extensions:
4
	VK_EXT_debug_report
	VK_EXT_debug_utils
	VK_KHR_portability_enumeration
	VK_LUNARG_direct_driver_loading
-9
failed to create instance!
Program ended with exit code: 1
```

if I'm using the variable from the `setup-env.sh` from the VulkanSDK

Screenshot 2023-12-04 at 22 23 26

```
required extensions:
3
	VK_KHR_surface
	VK_EXT_metal_surface
	VK_KHR_portability_enumeration
available extensions:
15
	VK_KHR_device_group_creation
	VK_KHR_external_fence_capabilities
	VK_KHR_external_memory_capabilities
	VK_KHR_external_semaphore_capabilities
	VK_KHR_get_physical_device_properties2
	VK_KHR_get_surface_capabilities2
	VK_KHR_surface
	VK_EXT_debug_report
	VK_EXT_debug_utils
	VK_EXT_metal_surface
	VK_EXT_surface_maintenance1
	VK_EXT_swapchain_colorspace
	VK_MVK_macos_surface
	VK_KHR_portability_enumeration
	VK_LUNARG_direct_driver_loading
Program ended with exit code: 0
```

Thoughts: 
- I was thinking about adding the whole list of environment variables from the `setup-env.sh` but not sure if needed.

### DIFF
--- a/en/02_Development_environment.adoc
+++ b/en/02_Development_environment.adoc
@@ -565,6 +565,7 @@ On Xcode toolbar go to `Product` > `Scheme` > `+Edit Scheme...+`, and in the `Ar
 
 * VK_ICD_FILENAMES = `vulkansdk/macOS/share/vulkan/icd.d/MoltenVK_icd.json`
 * VK_LAYER_PATH = `vulkansdk/macOS/share/vulkan/explicit_layer.d`
+* VK_DRIVER_FILES = `vulkansdk/macOS/share/vulkan/icd.d/MoltenVK_icd.json`
 
 It should look like so:
 


SaschaWillems:
Thank you very much for your PR.

Can you elaborate why this change is required?

0x401080:
@SaschaWillems thank you for checking it. I was going through the [official tutorial](https://docs.vulkan.org/tutorial/latest/02_Development_environment.html#_setting_up_xcode), when I met the issue, I'm using MacOS + Xcode, I was following the set up guide, then when I tried to run the compiled code for the [instance](https://docs.vulkan.org/tutorial/latest/03_Drawing_a_triangle/00_Setup/01_Instance.html#_creating_an_instance) section, there was an issue with the required and available extensions ( screenshots above ), I had set up env variables as it was in guide, but still I had the issue. 

I didn't install Vulkan globally, so mostly I rely on env variables, and it was like I was missing something to run it properly. 
I double check the official guide, and googled the issue, but it didn't help, I also had the makefile + VSCode, it didn't work either, but then I run `setup-env.sh` and I was able to compile the code via makefile, so I just narrowed down the env variables and added them to Xcode's env variables, I tried different combinations, and found out that `VK_DRIVERS_FILES` is working alone. But I'm not finished the tutorial yet, so maybe additional variables will be needed, so far the env variables from the official guide is not enough to run the code via Xcode, so I decided to create this PR. 


I hope it makes sense, please let me know, thank you

PR #40, opened 2024-04-14
Remove GLM_FORCE_RADIANS which is deprecated since glm 0.9.6.0 (2014-11-30)

clbrunet:
[GLM_FORCE_RADIANS deprecation](https://github.com/g-truc/glm?tab=readme-ov-file#deprecation-4)

This PR prevents people from copy-pasting the now useless GLM_FORCE_RADIANS definition.


SaschaWillems:
Thank you for this PR. Very much appreciated!

PR #47, opened 2024-07-20
Removed unnecessary variable

NecdetYavas:
We already have a class variable "swapChainExtent". Local "extent" variable seems unnecessary since we throw an error if swapchain creation fails.


NecdetYavas:
Forgot to edit the tutorial accordingly. Sorry for the inconvenience.

PR #66, opened 2025-06-01
GLSLC UBUNTU - Update 02_Development_environment.adoc

Nabir14:
I have tried installing glslc from my terminal on Ubuntu. It installed it without any errors and the programs compiled successfully. It's fast and simple.


SaschaWillems:
Not sure about this. Last time I checked the Ubuntu glslc package was outdated (glsl from 2023). and looked unmaintained. Did that change?

If at all, this should be using the slangc package, which includes glscl and seems to be actively maintained.

Nabir14:
> Not sure about this. Last time I checked the Ubuntu glslc package was outdated (glsl from 2023). and looked unmaintained. Did that change?
> 
> If at all, this should be using the slangc package, which includes glscl and seems to be actively maintained.

Hmm. I rechecked and it seems that it does use an outdated version from 2023:
```
shaderc 2023.2-1
spirv-tools 2022.4+1.3.236.0-1                          glslang 11.13.0-1

Target: SPIR-V 1.0
```

Well I guess I forgot about outdated problems with the debian repositories.

PR #69, opened 2025-07-01
Fix some more links

SaschaWillems:
@marty-johnson59 @gpx1000 : Hopefully these were the last broken links. Antor build now runs without issues for me.

We might want to merge as soon as possible.


marty-johnson59:
Thanks for finding all these!  Merging now - LMK if it causes problems.

PR #70, opened 2025-07-01
Fix chapter links

SaschaWillems:
@marty-johnson59 @gpx1000 : Sorry for the noise, but I find a few more links that were broken. And as they were using "link" instead of (the correct) "xref", they didn't show up as errors when building the site before.


gpx1000:
Thanks for this Sascha!  I wish I got these the first time.  That PR became really hard to find all the bugs in.

SaschaWillems:
Yeah, actually impossible to find in code and needs visual inspection.

But compared to how large the PR was these are very minor ;)

marty-johnson59:
Thanks guys.  Merging

PR #77, opened 2025-07-02
Set MSVC debugger working directory at CMake

SaschaWillems:
This adjusts the CMake build setup script to pass the correct debugger working directory to Visual Studio. With this change you can run samples out of the box, without having to adjust the working directory for each chapter by hand.


JoseEmilio-ARM:
Tested and works well on my side, thank you!

PR #78, opened 2025-07-02
Add layer and extension checks on instance creation

asuessenbach:
Note 1: you should mention the existence of the `install_dependencies` scripts in the readme.md!

Note 2: not all projects are successfully building. `00_base_code`, for example, fails.

### DIFF
--- a/attachments/01_instance_creation.cpp
+++ b/attachments/01_instance_creation.cpp
@@ -1,3 +1,4 @@
+#include <algorithm>
 #include <iostream>
 #include <stdexcept>
 #include <cstdlib>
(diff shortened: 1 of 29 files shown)


gpx1000:
Thanks @asuessenbach I'll add the mention to the readme and we have a fix for 00_base_code in the CI PR so fix for that is incoming.  This PR looks good to me, we'll need to work this change into all of the other code files as well that reproduces it.

asuessenbach:
> we'll need to work this change into all of the other code files as well that reproduces it.

I can extend this PR accordingly.

asuessenbach:
The function createInstance in 02_validation_layers.cpp doesn't need a fix, but just some checks.

Added them right now. If that looks ok, I can carry that over to the other sources.

gpx1000:
That looks great to me!  Thanks for the work!  I'm working on new chapters right now and have begun to enforce using vulkanconfig to set the validation layers so that change should only need to propagate through compute_shaders.  If that gets tedious, I'm happy to do the propagation.

SaschaWillems (approved):

Issue #75, opened 2025-07-01
Intellisense for Vulkan-hpp module not working with Visual Studio


SaschaWillems:
Intellisense for C++ Modules has always been hit-or-miss, but even with an up-to-date install, intellisense does not work at all. Code does compile, but working on code like that is impossible:

![Image](https://github.com/user-attachments/assets/c55faad6-994b-45f4-80da-e06168af13ff)

It looks like the vulkan-hpp module needs to be marked as a module at project level.

SaschaWillems:
So explicitly adding the module to each and every chapter's solution does make intellisense work. Compilation then fails as the solution also contains a separate project for the VulkanCpp-Module. Removing that project resolves that.

With these changes both building AND intellisense works for me. Not sure how to adjust CMake for this though.

PawKanarek:
I have the same issue in VS Code when using new c++ modules on macOS.

<img width="534" height="173" alt="Image" src="https://github.com/user-attachments/assets/76101a8d-5541-40d4-a3a8-c2c3979da455" />


i have added VukanCppModule to CMakeList, project compiles, but sadly no intelisense

```
add_library(VulkanCppModule)
add_library(Vulkan::cppm ALIAS VulkanCppModule)

target_compile_definitions(VulkanCppModule
        PUBLIC
        VULKAN_HPP_DISPATCH_LOADER_DYNAMIC=1
        VULKAN_HPP_NO_STRUCT_CONSTRUCTORS
)
target_include_directories(VulkanCppModule
        PRIVATE
        "${Vulkan_INCLUDE_DIRS}"
)
target_link_libraries(VulkanCppModule
        PUBLIC
        "${Vulkan_LIBRARIES}"
)

set_target_properties(VulkanCppModule PROPERTIES CXX_STANDARD 20)

target_sources(VulkanCppModule
        PUBLIC
        FILE_SET cxx_modules TYPE CXX_MODULES
        BASE_DIRS
        "${Vulkan_INCLUDE_DIRS}"
        FILES
        "${Vulkan_INCLUDE_DIRS}/vulkan/vulkan.cppm"
)
```

PawKanarek:
My workaround is to use neovim :D

<img width="789" height="493" alt="Image" src="https://github.com/user-attachments/assets/47175897-8c2f-4644-b01e-5dd6ee2fc18b" />


Finally found something that will push me deep into this rabbit hole without regrets :)

PR #83, opened 2025-07-05
Intellisense

gpx1000:
Included `vulkan_hpp_intellisense.h` across all attachment files to resolve C++20 module intellisense issues in IDEs.  This is a blind test to help fix: #75


SaschaWillems:
Sadly doesn't do anything for me with VS2022. Still no intellisense.

Only way I can get intellisense to work is remove the VulkanCppModule project and explicitly add the vulkan-hpp module to each project. Might be an option.

SaschaWillems:
Would love to do a PR myself, but my CMake knowledge simply isn't goot enough. I can't get it to properly work :/

gpx1000:
I tried a version with direct module linking, try turning that option on and see if it does the trick for you.  If it works, maybe we make it a default build method.

SaschaWillems:
Sorry for the stupid question, but how do I actually set/enable that option?

gpx1000:
No question is stupid; if you're using the command line just pass it in to cmake like this:
cmake -d<OPTION_NAME>=<VALUE> ..
or:
mkdir build && cd build
cmake -dUSE_DIRECT_MODULE_IMPORT=ON ..

gpx1000:
I also left comments in the cmake to detail how to set the option.

SaschaWillems:
Didn't seem to change anything for me on VS2022. Will try a few different things first and get back to you.

gpx1000:
okay, this version I tried to directly tell intellisense that it needs to include vulkan.hpp directly.
#ifdef __INTELLISENSE__
#include <vulkan/vulkan.hpp>
#else
import vulkan_hpp;
#endif
https://stackoverflow.com/questions/26724244/is-there-a-macro-defined-when-intellisense-is-parsing  that macro should hopefully be the key to getting intellisense to behave.

SaschaWillems:
That does give me Intellisense, but only partially:

image


Stuff inside the RAII namespace (?) still doesn't get resolved.

But does that still use the module then, or just the normal header?

gpx1000:
Yep, it still uses the module, only when INTELLISENSE is running (i.e. the static analysis step of "compiling" the source file) will it include the whole vulkan hpp file.  Rest of the time it should be the module.  Think I need to include vulkan_raii.hpp to get the right stuff instead of vulkan hpp by itself.  I hate doing this, macros for a static analysis platform are silly, detract from understanding the code and don't precisely give what should exist from the compiler which is the whole reason static analysis exists.  But I'll get off my soapbox as I'm sure Intellisense will eventually work with modules sooner or later.

gpx1000:
Okay, all of the tutorial should now resolve correctly under intellisense.  Lemme know if I missed anything.

SaschaWillems:
That did it. Intellisense is now properly working for me 👍🏻

Issue #73, opened 2025-07-01
Resizing no longer working


SaschaWillems:
On my setup (Win11, NV RTX 4070, latest Vulkan developer driver), all samples will exit with "" when closing:

```
vk::Queue::presentKHR: ErrorOutOfDateKHR

V:\...\Vulkan-Tutorial\attachments\build\30_multisampling\Debug\30_multisampling.exe (process 19868) exited with code 1 (0x1).
```

It looks like this is an exception thrown by the vulkan-hpp module, which is not handled by the app, and as such results in the app exiting.

SaschaWillems:
Any update on this? I think it's one of the most pressing issues as this affects anyone, at least on windows.

gpx1000:
Yeah, I'm doing a bit more work on it tonight.  Trying to get a few changes finished for the game engine tutorial will switch over as soon as I can.

SaschaWillems:
Any update on this? If you can't find the time to fix, I might take a look at it.

gpx1000:
I haven't had a chance to work on it yet.  I'll try to look at it today

SaschaWillems:
Can we get this fixed? This also happens on the new ray tracing chapters, making it crash on resize.

nulface:
When running Swap chain recreation,
When resizing window:

```
validation layer: type { Validation } msg: vkDestroyFence(): can't be called on VkFence 0x140000000014 that is currently in use by VkQueue 0x46034d0.
The Vulkan spec states: All queue submission commands that refer to fence must have completed execution (https://vulkan.lunarg.com/doc/view/1.4.313.2/windows/antora/spec/latest/chapters/synchronization.html#VUID-vkDestroyFence-fence-01120)
validation layer: type { Validation } msg: vkDestroyFence(): can't be called on VkFence 0x150000000015 that is currently in use by VkQueue 0x46034d0.
The Vulkan spec states: All queue submission commands that refer to fence must have completed execution (https://vulkan.lunarg.com/doc/view/1.4.313.2/windows/antora/spec/latest/chapters/synchronization.html#VUID-vkDestroyFence-fence-01120)
validation layer: type { Validation } msg: vkDestroySemaphore(): can't be called on VkSemaphore 0xf000000000f that is currently in use by VkQueue 0x46034d0.
[...]

PR #85, opened 2025-07-05
Add exception handling for vk::SystemError

gpx1000:
This is specific to the issue #73 

- Catch `vk::SystemError` in `mainLoop` and presentation logic to handle `vk::Result::eErrorOutOfDateKHR` gracefully.
- Log and ignore swapchain-related errors during shutdown or presentation to prevent crashes while rethrowing unexpected exceptions.
- Updated multiple attachments for consistent error handling.

We likely will need to add better error detection everywhere; but in the spirit of just handling it when we get an out of date.  It should be properly handled already in the draw function; but this makes it so we don't just throw an exception.  I guess a better solution might be to handle the `vk::Result::eErrorOutOfDateKHR` by sending it to resize multiple times rather than the once.  However, I think this is the right solution for now for this.

### DIFF
--- a/attachments/15_hello_triangle.cpp
+++ b/attachments/15_hello_triangle.cpp
@@ -100,7 +100,18 @@ private:
     void mainLoop() {
         while (!glfwWindowShouldClose(window)) {
             glfwPollEvents();
-            drawFrame();
+            try {
+                drawFrame();
+            } catch (const vk::SystemError& e) {
+                if (e.code().value() == static_cast<int>(vk::Result::eErrorOutOfDateKHR)) {
+                    // Swapchain is out of date, this can happen during window close
+                    // Just ignore and continue to close
+                    std::cout << "Ignoring ErrorOutOfDateKHR during shutdown" << std::endl;
+                } else {
+                    // Rethrow other errors
+                    throw;
+                }
+            }
         }
 
         device.waitIdle();
--- a/attachments/16_frames_in_flight.cpp
+++ b/attachments/16_frames_in_flight.cpp
@@ -104,7 +104,18 @@ private:
     void mainLoop() {
         while (!glfwWindowShouldClose(window)) {
             glfwPollEvents();
-            drawFrame();
+            try {
+                drawFrame();
+            } catch (const vk::SystemError& e) {
+                if (e.code().value() == static_cast<int>(vk::Result::eErrorOutOfDateKHR)) {
+                    // Swapchain is out of date, this can happen during window close
+                    // Just ignore and continue to close
+                    std::cout << "Ignoring ErrorOutOfDateKHR during shutdown" << std::endl;
+                } else {
+                    // Rethrow other errors
+                    throw;
+                }
+            }
         }
 
         device.waitIdle();
--- a/attachments/17_swap_chain_recreation.cpp
+++ b/attachments/17_swap_chain_recreation.cpp
@@ -111,7 +111,18 @@ private:
     void mainLoop() {
         while (!glfwWindowShouldClose(window)) {
             glfwPollEvents();
-            drawFrame();
+            try {
+                drawFrame();
+            } catch (const vk::SystemError& e) {
+                if (e.code().value() == static_cast<int>(vk::Result::eErrorOutOfDateKHR)) {
+                    // Swapchain is out of date, this can happen during window close
+                    // Just ignore and continue to close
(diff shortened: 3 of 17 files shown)


SaschaWillems (changes requested):
This change now results in an infinite loop for me when trying to resize the  application:

image

That message is spammed endlessly, the application no longer reacts.

gpx1000:
I *just* tried to fix 17.  @SaschaWillems, do me a favor and try that swapchain resize chapter code in windows and see if it fixes resize.  I'll bring it to all of the other files if it solves it.

SaschaWillems:
No change:

https://github.com/user-attachments/assets/cd4eb0da-5fb5-471f-ae4e-119e418025e2

I don't see proper resize handling in that chapter (or any other chapter).

gpx1000:
I think there's a difference in our systems.  Debugging in Linux, I'm seeing it correctly return the result of outofdate and does the recreate swapchain thing.  However, you're getting spam, which means that it is throwing that error.  Thus the "right" thing to do I think is to call the recreate swapchain in the catch block instead of just printing out to console.  I mean that makes sense to me that it would rethrow each frame instead of sending out the error message which is admittedly what I'm used to.
Sorry about asking you to try again, but I think I might have this fixed again, we're only trying out chapter 17 until you say we're good.

SaschaWillems:
The latest version of chapter 17 does work indeed. No more issues, so feel free to apply the change to all chapters 👍🏻

gpx1000:
Okay, I think Resizing is hopefully working for all.

gpx1000:
Okay, all of the chapters should handle resizing correctly.

attachments/34_android.cpp
```diff
@@ -1281,47 +1281,69 @@ class HelloTriangleApplication {
...
+        if (window) {
+            glfwGetFramebufferSize(window, &width, &height);
+            while (width == 0 || height == 0) {
+                glfwGetFramebufferSize(window, &width, &height);
+                glfwWaitEvents();
+            }
+        }
+#endif
         // Wait for device to finish operations
         device.waitIdle();
```
cubanismo, line 1337: Note per recent spec updates, waiting for idle on the queue used for presentation is sufficient.

Also, it would be best if VK_EXT_swapchain_maintenance1's present fence were used here instead.

Is there a tutorial yet for smooth resize? I.e., using oldSwapchain to create the new swapchain and deferring cleanup until pending swaps have reached a point where it's safe to delete the old sawpchain(s)?

attachments/33_vulkan_profiles.cpp
```diff
@@ -1548,28 +1582,29 @@ class HelloTriangleApplication {
         };
         queue.submit(submitInfo, *inFlightFences[currentFrame]);
 
-        const vk::PresentInfoKHR presentInfoKHR{
-            .waitSemaphoreCount = 1,
-            .pWaitSemaphores = &*presentCompleteSemaphore[imageIndex],
```
cubanismo, line 1554: Why does this example still have renderFinishedSemaphores, but also have these presentCompleteSemaphore array that seems to server the same purpose, but with a slightly misleading name (It seems to be used more as a "rendering complete/present-ready" indicator, rather than a "present complete" indicator, the latter of which I'm not sure is possible to measure using semaphores in current Vulkan, which is what lead me to look closely at this code).

The Android file below seems to still use renderFinishedSemaphores in the manner that fits its name for what appears to be the same steps renderFinishedSemaphores is used here.

attachments/34_android.cpp
gpx1000: Yep, we should open 3 issues you presented here:
1.) remove not needed device.waitIdle();
2.) use swapchain_maintenance1
3.) request to create a smooth resize tutorial.

The first two will require making changes to the tutorial itself as we don't want to just change the code without making sure the tutorial explains what needs to be done to be correct.  The last will require a full tutorial which takes more time and planning; however, it should be done as we do see the request from time to time to give a sample on smooth resizing.

attachments/33_vulkan_profiles.cpp
gpx1000: 'opps' is the why :)  I made a bug.  Thanks for the report, I'll address it.

attachments/34_android.cpp
SaschaWillems: We also need to take into account that VK_EXT_swapchain_maintenance1 isn't universally supported, and the KHR variant even less so. So would prob. need to be optional for now.

PR #99, opened 2025-07-16
Remove deprecated `Findtinygltf.cmake` file

gpx1000:


SaschaWillems:
I've also been having the issue that @asuessenbach mentioned regarding this file. I'm on Windows, and for whatever reason that FindTinyGLTF.cmake does some very weird things so that I always have to delete and fresh clone the repo.

To test this PR, all I did:

- Check out a clean copy of the repo
- Check out this PR

And just by doing that I get the following diff:

image

That feels very odd/wrong. Not sure what's causing this on Windows, but also not sure how to tell if this actually fixes the issue.

gpx1000:
hmm...  maybe we need to rename it to the lower case and drop the camel case.  Windows + fetch_content...  strange things happening.  Not sure what's causing it.
Of course the Linux side is failing in CI because something changed in the latest SDK.  So, I have a few things to fix here.

SaschaWillems:
Yeah. weird indeed. If I do a 'git reset --hard', the change disappears. But then git switch to main and git switch back to this PR magically makes that diff appear again :/

asuessenbach:
The actual issue here is, that Windows is inherently case insensitive.

I would assume that, if one of the two Find*.cmake files is gone and you freshly clone this repo, every problem around that file should be gone.
Maybe, @gpx1000, it's worth to just remove that file, ignoring all your other CI issues you currently have (for now), and verify that it really helps.

gpx1000:
Still not entirely sure what exactly was wrong, but combo of reverting, forcing the naming to be correct and getting gltf to be happy seems to have fixed CI.  Please give this a try and see if it solves your problems in Windows.

SaschaWillems:
Still the same issue when doing a clean checkout with main: As soon as I switch to any branch (including this), it has the issue with the Findtinygltf.cmake file name.

I think we need to merge in order to properly test this in Windows as the issue is in main.

gpx1000:
Okay, merging, it can't be worse than it currently is ;-)

SaschaWillems:
Looking good. Just did a clean checkout, switched to a branch without any issues.

PR #101, opened 2025-07-19
cleanupSwapChain in 17_swap_chain_recreation.cpp was not destroying the previous swapchain

lmichaudel:
As the title says, the source code was missing a simple `swapChain = nullptr;` (that is actually present in the tutorial).
Without this fix the swapChain recreation crashes (`Device::createSwapchainKHR: ErrorSurfaceLostKHR`).

### DIFF
--- a/attachments/17_swap_chain_recreation.cpp
+++ b/attachments/17_swap_chain_recreation.cpp
@@ -124,6 +124,7 @@ private:
 
     void cleanupSwapChain() {
         swapChainImageViews.clear();
+        swapChain = nullptr;
     }
 
     void cleanup() {


gpx1000:
Great find!  This might be the explanation behind the resize crash issue in Windows.  @SaschaWillems if you have a chance, could you verify if this brilliant find solves resize?

lmichaudel:
I checked all the other cleanup methods and fixed them. 
`33_vulkan_profiles.cpp` was actually already fixed.
EDIT: `34_android.cpp` too.

gpx1000:
Well it's a great find, you're absolutely right those should have been cleaned up.  Don't worry about CI failing.  That has been having problems; I'm more testing if CI passes Windows.

gpx1000:
If you sync to the latest then the CI should pass for all build environments.

SaschaWillems:
Still crashing for me. This does not address the actual issue, which is this block of code:

```cpp
                result = presentQueue.presentKHR(presentInfo);
                if (result == vk::Result::eErrorOutOfDateKHR || result == vk::Result::eSuboptimalKHR || framebufferResized) {
                    framebufferResized = false;
                    recreateSwapChain();
                }
                else if (result != vk::Result::eSuccess) {
                    throw std::runtime_error("failed to present swap chain image!");
                }
```

presentQueue.presentKHR will throw an exception, but there is no exception handing around it.

gpx1000:
Well, that's sad, I'll leave fixing it on my TODO list.  For now, this PR is correct and necessary.  LGTM.

Issue #126, opened 2025-08-04
Code identation and line breaks are all over the place


SaschaWillems:
Looking and some of the code files, I noticed that they're sometimes very hard to read as they lack consistent identation and line break styles.

E.g. in the compute shader chapter we have stuff like this:

```cpp
        vk::ImageMemoryBarrier2 barrier = {
            .srcStageMask = src_stage_mask,
            .srcAccessMask = src_access_mask,
            .dstStageMask = dst_stage_mask,
            .dstAccessMask = dst_access_mask,
            .oldLayout = old_layout,
            .newLayout = new_layout,
            .srcQueueFamilyIndex = VK_QUEUE_FAMILY_IGNORED,
            .dstQueueFamilyIndex = VK_QUEUE_FAMILY_IGNORED,
            .image = swapChainImages[imageIndex],
            .subresourceRange = {
                .aspectMask = vk::ImageAspectFlagBits::eColor,
                .baseMipLevel = 0,
                .levelCount = 1,
                .baseArrayLayer = 0,
                .layerCount = 1
            }
        };
```

Properly aligned in indented. Easy on the eyes.

But just a few lines later we have this:

```cpp
        const vk::SubmitInfo submitInfo{ .waitSemaphoreCount = 1, .pWaitSemaphores = &*presentCompleteSemaphore[semaphoreIndex],
                            .pWaitDstStageMask = &waitDestinationStageMask, .commandBufferCount = 1, .pCommandBuffers = &*commandBuffers[currentFrame],
                            .signalSemaphoreCount = 1, .pSignalSemaphores = &*renderFinishedSemaphore[imageIndex] };
```

Odd identation, no line breaks, almost unreadable.

And a few lines down:

```
vk::raii::ShaderModule shaderModule{ device, createInfo };
```

No line breaks, no designated initializers.

In general, some of the code files are very hard to read due to a lack of consistent code style.

SaschaWillems:
Maybe we should introduce a clang format file like we did for the Vulkan samples.

SaschaWillems:
We did introduce clang format, so closing this.

PR #131, opened 2025-08-06
Add clang format from samples repo

SaschaWillems:
As noted in #126 code style is all over the place, making things hard to read.

We should start using a clang format file like we did in the samples to make sure code style follows a fixed set of guidelines.

This PR is there to discuss that. It adds the clang format file from the samples and applies that to one chapter.

While modern C++ is generally hard on the eyes, this makes it at least somewhat easier to read and (more importantly) more consistent.

### DIFF
--- a/attachments/.clang-format
+++ b/attachments/.clang-format
@@ -1,0 +1,111 @@
+---
+Language:        Cpp
+# BasedOnStyle:  LLVM
+AccessModifierOffset: -2
+AlignAfterOpenBracket: Align
+AlignConsecutiveAssignments: true
+AlignConsecutiveDeclarations: true
+AlignEscapedNewlines: Left
+AlignOperands:   true
+AlignTrailingComments: true
+AllowAllParametersOfDeclarationOnNextLine: true
+AllowShortBlocksOnASingleLine: false
+AllowShortCaseLabelsOnASingleLine: false
+AllowShortFunctionsOnASingleLine: None
+AllowShortIfStatementsOnASingleLine: false
+AllowShortLoopsOnASingleLine: false
+AlwaysBreakAfterDefinitionReturnType: None
+AlwaysBreakAfterReturnType: None
+AlwaysBreakBeforeMultilineStrings: false
+AlwaysBreakTemplateDeclarations: true
+BinPackArguments: true
+BinPackParameters: true
+BraceWrapping:
+  AfterCaseLabel: true
+  AfterClass:      true
+  AfterControlStatement: true
+  AfterEnum:       true
+  AfterFunction:   true
+  AfterNamespace:  true
+  AfterObjCDeclaration: true
+  AfterStruct:     true
+  AfterUnion:      true
+  AfterExternBlock: true
+  BeforeCatch:     true
+  BeforeElse:      true
+  IndentBraces:    false
+  SplitEmptyFunction: false
+  SplitEmptyRecord: false
+  SplitEmptyNamespace: false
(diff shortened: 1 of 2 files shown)


asuessenbach:
Besides lots of other settings I'm using with the Vulkan-Hpp project, something like
    `ColumnLimit : 160`
helps very much in getting some nice formatting, preventing arbitrary long lines.

Just for reference, I've attached the clang-format from Vulkan-Hpp as a text file: [clang-format.txt](https://github.com/user-attachments/files/21660232/clang-format.txt)

SaschaWillems:
As agreed on the last call: I'll apply the clang format to all files and update this PR.

SaschaWillems:
No idea why the Android build fails.

asuessenbach:
> No idea why the Android build fails.

clang_format has sorted the includes. Now, `vulkan_android.h` is included before `vulkan_core.h` in `34_android.cpp`.

SaschaWillems:
Oh, that was prob. stupid. Will redo this PR without reordering the includes :/

asuessenbach:
> Oh, that was prob. stupid.

No, that was not stupid. It's vulkan, that is stupid.
I would prefer to just exclude those two `#includes` from reordering/formatting.

Or maybe just `#include <vulkan/vulkan.h>`, instead.

SaschaWillems:
That would be the better option. Didn't realize that the tutorial did explicitly include the core/platform specific headers. That indeed does not make much sense and should be changed indeed. Also feels odd that every chapter explicitly includes "vk_platform.h".

gpx1000 (approved):
looks like merge conflicts.  Could you address?

attachments/.clang-format
```diff
@@ -86,7 +86,7 @@ PenaltyExcessCharacter: 1000000
 PenaltyReturnTypeOnItsOwnLine: 60
 PointerAlignment: Right
 ReflowComments:  true
-SortIncludes:    true
+SortIncludes:    false
```
asuessenbach: I thought, you wanted to keep the includes sorted, just handle the android cases locally by either just including `vulkan.hpp`, or not sorting those headers by using `//clang-format off` and `//clang-format on`

SaschaWillems: Yeah, but if I enable this globally, there is a chance a maintainer forgets about that. Will try with the //clang-format off route (didn't know that existed 👍🏻)

gpx1000:
Probably a good idea to add it to CI in same way we do in Samples so it fails if maintainers forget ;-).

SaschaWillems:
Having CI builds for all supported platforms would be very helpful indeed.

gpx1000:
Well we already have CI build for all platforms, this should just be the same copyright/clang check that Samples uses.  https://github.com/KhronosGroup/Vulkan-Samples/blob/main/.github/workflows/check.yml  take that and update pathing for here should *just work.* hopefully?  Maybe add the ignore stuff as in what we use there.

SaschaWillems:
I have re-enabled include ordering (and toggle it off for the few android includes we have) and have (re)applied Clang Format to all source files. Also added a note on ClangFormat to the contribution guide.

The earlier we can merge this, the easier things get, esp. for maintainers ;)

SaschaWillems:
@marty-johnson59 , @gpx1000 : I have fixed the merge conflicts and reapplied clang-format to all files. From my perspective, this is now ready to be merged as discussed on the last docs call.

gpx1000 (approved):

SaschaWillems:
Thx :)

PR #149, opened 2025-08-18
Remove explicit CMake version requirement

SaschaWillems:
This PR removes the explicit CMake version requirement for Android builds. It was pinned at 4.02, so everyone with a different version would see an error that CMake 4.0.2 could not be found. Even if you have a newer version installed.

Having that explicit requirement pinned to a single version is not necessary and as far as I can tell is safe to remove.

This PR makes it a min. requirement instead.

### DIFF
--- a/attachments/android/app/build.gradle
+++ b/attachments/android/app/build.gradle
@@ -38,6 +38,7 @@ android {
     externalNativeBuild {
         cmake {
             path "src/main/cpp/CMakeLists.txt"
+            version "4.0.2+"
         }
     }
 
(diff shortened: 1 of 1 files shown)


SaschaWillems:
CI oddly fails. For whatever reason it tries to use CMake 3.22.1.

gpx1000:
Maybe try replacing the version = '4.02+'  The + used to allow gradle to select any newer version.  I think that's frowned upon as a practice; but it might still work?

SaschaWillems:
That's what I did. That fixed the CI, and works fine for me using a more recent version (that 4.0.2) 👍🏻

PR #155, opened 2025-08-25
Replace defensive clear with assert in createImageViews

asuessenbach:
I don't see, how `swapChainImageViews` could be not empty in `createImageViews()`. Do I miss something?

In any case, I would replace the defensive `clear()` with an `assert`, to make clear it's a prerequisite for this function.

### DIFF
--- a/attachments/33_vulkan_profiles.cpp
+++ b/attachments/33_vulkan_profiles.cpp
@@ -237,6 +237,7 @@ private:
             imageView = nullptr;
         }
 
+        swapChainImageViews.clear();
         swapChain = nullptr;
     }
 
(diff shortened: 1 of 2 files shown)


SaschaWillems:
Does make sense in this chapter. For msvc this needs to include assert though.

Looking at other chapters, the createImageViews function needs some fix ups in other chapters too. In some of the chapters (e.g. 23) it just emplaces back with no clear, so every resize will just add new image views.

asuessenbach:
Asserting on empty `swapChainImageViews` in `createImageViews()` in all chapters.

Chapters supporting resizing already clear the `swapChainImageViews` in `cleanupSwapChain()`

Besides that, all chapters staring with 17_swap_chain_recreation are crashing on resize.

SaschaWillems:
Thanks 👍🏻 

Crash on resize will be fixed in #85

We should merge that as soon as possible

PR #156, opened 2025-08-25
Fix usage of backquotes with word characters on either side.

oddhack:
Fixes https://github.com/KhronosGroup/Vulkan-Guide/issues/212 (when combined with https://github.com/KhronosGroup/Vulkan-Guide/pull/342).

Note that this renders as (tt font) VkImage closely followed by (regular font) s, which is fine, but it might be more clear to write something like "the VkImage objects/handles".

### DIFF
--- a/en/03_Drawing_a_triangle/01_Presentation/01_Swap_chain.adoc
+++ b/en/03_Drawing_a_triangle/01_Presentation/01_Swap_chain.adoc
@@ -505,7 +505,7 @@ image::/images/swap_chain_validation_layer.png[]
 
 == Retrieving the swap chain images
 The swap chain has been created now, so all that remains is retrieving the
-handles of the ``VkImage``s in it. We'll reference these during rendering
+handles of the `VkImage` objects it contains. We'll reference these during rendering
 operations in later chapters.
 
 [,c++]


en/03_Drawing_a_triangle/01_Presentation/01_Swap_chain.adoc
```diff
@@ -505,7 +505,7 @@ image::/images/swap_chain_validation_layer.png[]
 
 == Retrieving the swap chain images
 The swap chain has been created now, so all that remains is retrieving the
-handles of the `VkImage`s in it. We'll reference these during rendering
+handles of the ``VkImage``s in it. We'll reference these during rendering
```
oddhack:
```suggestion
handles of the `VkImage` objects in the swap chain. We'll reference these during rendering
```

for example.

MarijnS95: Mentioning `swap chain` again in the same sentence feels a bit superfluous, but otherwise makes sense :)

oddhack: Could also say "the VkImage objects it contains". It is annoying when our writing is constrained by the limits of our markup language :-(

MarijnS95: Yep, the additional spacing and font change make this less pretty/clear to read. I wonder what the alternatives are.

EDIT: Anyway, the original concern/issue were about the backtick visibly rendering between the word and the `s`.

oddhack: I applied my suggested rephrasing to avoid the ugly markup. Hopefully that's OK with everyone.

oddhack:
@marty-johnson59 @SaschaWillems this is ready to be merged - if you agree, please go ahead.

marty-johnson59 (approved):

gpx1000 (approved):

PR #157, opened 2025-08-26
docs: removed wrong information

CyntexMore:
* removed block containing "...we’ll be using the original C API..."
* capitalized "RAII"


marty-johnson59:
Merging - per 8 September discussion

PR #161, opened 2025-09-04
Fixed code snippet in Update 00_Base_code.adoc

rapfamily4:
There was an example Update 00_Base_code.adoc that wouldn't compile.


SaschaWillems:
The fix is fine, but this code is not used anywhere in the tutorial's code. Might be better to remove or adjust to what is actually used in code. Any thoughts @gpx1000 ?

gpx1000:
Yeah, should try to update it to what the code does.

SaschaWillems:
Will be addressed by #226 226

PR #166, opened 2025-09-17
Fix: Set queue priority to 1.0f in all samples and docs

gavrix32:
Replaced all instances of:
```cpp
float queuePriority = 0.0f;
```
with:
```cpp
float queuePriority = 1.0f;
```
Updated in both `.adoc` and `.cpp` files (33 total), to reflect best practice.

`1.0f` is the recommended queue priority, even in single-queue setups, to avoid confusion for beginners and ensure optimal execution priority.

### DIFF
--- a/attachments/04_logical_device.cpp
+++ b/attachments/04_logical_device.cpp
@@ -211,7 +211,7 @@ class HelloTriangleApplication
 		};
 
 		// create a Device
-		float                     queuePriority = 0.0f;
+		float                     queuePriority = 0.5f;
 		vk::DeviceQueueCreateInfo deviceQueueCreateInfo{.queueFamilyIndex = graphicsIndex, .queueCount = 1, .pQueuePriorities = &queuePriority};
 		vk::DeviceCreateInfo      deviceCreateInfo{.pNext                   = &featureChain.get<vk::PhysicalDeviceFeatures2>(),
 		                                           .queueCreateInfoCount    = 1,
--- a/attachments/05_window_surface.cpp
+++ b/attachments/05_window_surface.cpp
@@ -229,7 +229,7 @@ class HelloTriangleApplication
 		};
 
 		// create a Device
-		float                     queuePriority = 0.0f;
+		float                     queuePriority = 0.5f;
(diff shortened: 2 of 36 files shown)


SaschaWillems:
> 1.0f is the recommended queue priority, even in single-queue setups, to avoid confusion for beginners and ensure optimal execution priority.

Where is that coming from? Do you have sources for this?

gavrix32:
> > 1.0f is the recommended queue priority, even in single-queue setups, to avoid confusion for beginners and ensure optimal execution priority.
> 
> Where is that coming from? Do you have sources for this?

The use of `1.0f` is consistent with the [official Khronos Hello Triangle sample](https://github.com/KhronosGroup/Vulkan-Samples/blob/main/samples/api/hello_triangle/hello_triangle.cpp), which sets full priority even for a single queue.

The [Vulkan specification](https://registry.khronos.org/vulkan/specs/latest/html/vkspec.html) (section 5.3.4 – Queue Priority) defines `1.0f` as the highest priority and `0.0f` as the lowest. Using `1.0f` avoids any unintended deprioritization and better reflects developer intent.

SaschaWillems:
>  Using 1.0f avoids any unintended deprioritization and better reflects developer intent.

I have never heard of that nor did I ever experience that. What is that "deprioritization" part based on?

gavrix32:
> > Using 1.0f avoids any unintended deprioritization and better reflects developer intent.
> 
> I have never heard of that nor did I ever experience that. What is that "deprioritization" part based on?

It's not based on a known issue — just on the spec wording that higher-priority queues may get more time. I used “deprioritization” in the sense of avoiding future ambiguity if other queues were added. It’s mostly about explicit intent and clearer code, especially in tutorials.

SaschaWillems:
We agreed on using 0.5f instead for the samples and will change those. Can you update your PR accordingly?

gavrix32:
> We agreed on using 0.5f instead for the samples and will change those. Can you update your PR accordingly?

Updated to 0.5f as requested. Thanks for the feedback!

SaschaWillems:
Looks good to me.

We just merged a larger PR that reformats code, so in order to merge this you need to fix the merge conflicts.

SaschaWillems (approved):

PR #167, opened 2025-09-17
Add missing query for support of shaderDrawParameters

asuessenbach:
Resolves a ValidationLayer error.


SaschaWillems:
Looks correct to me.

Issue #158, opened 2025-08-28
Should clarify that the `setup-env` script is linux only.


charles-lunarg:
https://github.com/KhronosGroup/Vulkan-Tutorial/blob/b65ff3258e129140af5fc4aeb66ea05089984150/en/02_Development_environment.adoc?plain=1#L68

This block of text is linux specific (windows SDK installer sets up the env-vars automatically, macOS has different patterns but also has an installer).

SaschaWillems:
Fixed via #176

PR #176, opened 2025-09-22
Clarify that setup.env is linux-only

johannes-el:
Minor change in wording on the section and clarification that `setup-env` is Linux-only as pointed out in #158.


SaschaWillems:
Thx for your PR. Very much appreciated.

Issue #134, opened 2025-08-06
All chapters starting with hello triangle have validation errors


SaschaWillems:
Running any of the chapters starting with hello triangle (didn't check earlier ones as they don't display anything meaningful) show a lot of validation errors.

This is an excerpt:

```
Validation Error: [ VUID-VkShaderModuleCreateInfo-pCode-08740 ] | MessageID = 0x6e224e9
vkCreateShaderModule(): SPIR-V Capability DrawParameters was declared, but one of the following requirements is required (VkPhysicalDeviceVulkan11Features::shaderDrawParameters OR VK_KHR_shader_draw_parameters).
The Vulkan spec states: If pCode is a pointer to SPIR-V code, and pCode declares any of the capabilities listed in the SPIR-V Environment appendix, one of the corresponding requirements must be satisfied (https://vulkan.lunarg.com/doc/view/1.4.321.1/windows/antora/spec/latest/chapters/shaders.html#VUID-VkShaderModuleCreateInfo-pCode-08740)

validation layer: type { Validation } msg: vkCreateShaderModule(): SPIR-V Capability DrawParameters was declared, but one of the following requirements is required (VkPhysicalDeviceVulkan11Features::shaderDrawParameters OR VK_KHR_shader_draw_parameters).
The Vulkan spec states: If pCode is a pointer to SPIR-V code, and pCode declares any of the capabilities listed in the SPIR-V Environment appendix, one of the corresponding requirements must be satisfied (https://vulkan.lunarg.com/doc/view/1.4.321.1/windows/antora/spec/latest/chapters/shaders.html#VUID-VkShaderModuleCreateInfo-pCode-08740)
Validation Error: [ SYNC-HAZARD-WRITE-AFTER-READ ] | MessageID = 0x376bc9df
vkQueueSubmit(): WRITE_AFTER_READ hazard detected. vkCmdPipelineBarrier2 (from VkCommandBuffer 0x1ff805ad3d0 submitted on the current VkQueue 0x1ff8025d900) writes to VkImage 0x40000000004, which was previously read by vkAcquireNextImageKHR.
The current synchronization defines the destination stage mask as VK_PIPELINE_STAGE_2_COLOR_ATTACHMENT_OUTPUT_BIT, but layout transition does not synchronize with these stages.
[...]

PR #178, opened 2025-09-29
Use vk::StructureChain on pipeline creation.

asuessenbach:
In chapter 12 and following, a structure chain is manually created on creating the pipeline.

Should we change that to using a vk::StructureChain, instead?

### DIFF
--- a/attachments/12_graphics_pipeline_complete.cpp
+++ b/attachments/12_graphics_pipeline_complete.cpp
@@ -326,15 +326,22 @@ private:
 
         pipelineLayout = vk::raii::PipelineLayout( device, pipelineLayoutInfo );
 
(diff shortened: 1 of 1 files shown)


SaschaWillems (approved):
This improves readability by a lot!

So yes, I'm for doing it like this in all subsequent chapters.

asuessenbach:
Especially have a look at chapters 32 and 33, which need some special handling on the vk::StructureChain.

Note that some chapters have errors both with and without this PR:
- 18, 22: validation layer errors, nothing is displayed
- 26, 38: validation layer warnings
- 35, 36: does not find some ktx.dll, exits

SaschaWillems:
The issues with 35, 36 are prob. related to vcpkg. They work fine for me, but getting those to work on Windows was quite a hassle :/

SaschaWillems:
> Note that some chapters have errors both with and without this PR:

Last time I checked pretty much all chapters had validation errors, we need to tackle these, see #134

Issue #190, opened 2025-10-05
Update overview chapter to reworked tutorial


SaschaWillems:
The whole overview chapter is still pretty much the same as it was before we reworked the tutorial. We need to update it to avoid confusion.

Some things that we need to adjust:
- Replace renderpasses with dynamic rendering
- Remove framebuffers
- Remove coupling of pipeline and renderpasses
- Replace C-style Vulkan calls with C++-style Vulkan calls

Refs #168

marty-johnson59:
PR 198 addresses this - closing

PR #193, opened 2025-10-05
added explanation for dynamic rendering

### LINKED ISSUE #168: Why does the overview section still cover Renderpasses?
So in the overview section of the tutorial Renderpasses are described.
But since the tutorial later on only uses dynamic rendering, I would suggest updating this part.

johannes-el:
This PR tries to resolve #168 by introducing an explanation for `dynamic rendering` in the overview section of the tutorial.

### DIFF
--- a/en/01_Overview.adoc
+++ b/en/01_Overview.adoc
@@ -133,16 +133,33 @@ Because there could be many different images in the swap chain, we'll
 preemptively create an image view and framebuffer for each of them and
 select the right one at draw time.
 
-=== Step 5 - Render passes
+=== Step 5 - Dynamic rendering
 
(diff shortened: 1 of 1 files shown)


SaschaWillems:
See #190, we should redo the whole chapter.

johannes-el:
@SaschaWillems Yes you are right. But I thought doing it in little pieces is a better way to resolving this.
I can work on the other points mentioned in #190, sure.

gpx1000:
Thanks for doing the dynamic rendering portion.  I didn't want to lose your changes while we work on the other parts of #190.  If you get to updating the other parts before we do, then we'll be grateful for any changes you suggest there.

PR #202, opened 2025-10-17
Add C language examples alongside C++ in Vulkan tutorials

gpx1000:
- Include equivalent C code snippets for Vulkan-related examples to enhance tutorial compatibility.
- Annotate source blocks to specify programming languages for clarity (`[source,multilang,c++,c]`).
- Improve cross-language consistency in method signatures, helper functions, and setup instructions.

Issue #122, opened 2025-08-04
Don't mix up Vulkan C header and CPP header syntax


SaschaWillems:
Already noted in some of the issues here, but also something that was raised on social channels like reddit and discord: The tutorial now mixes Vulkan C header syntax with Vulkan CPP header syntax. This seems to be confusing to readers, as often you have Vulkan C header syntax in the text for a chapter, but the actual code for the chapter uses the Vulkan CPP header syntax.

We should use one of those exclusively and not mix them up.

SaschaWillems:
We probably should remove/rephrase this:

```
To make it easier to understand the core concepts and to follow along for
developers using other programming languages, also to get some experience with
the base API we'll be using the original C API when we describe the objects
and the concepts that are being used.
```
(Introduction chapter)

That's confusing to people, as it sounds like the tutorial uses the C API in addition to the C++ API, which is not the case.

SaschaWillems:
This is coming up regularly on Discord and/or reddit. We should address this with a high priority.

sharadhr:
Happy to jump in with improved documentation on Vulkan-Hpp. I already have something in the works for the RAII headers, which haven't yet been updated.

PR #203, opened 2025-10-18
Improve Image Views chapter by adding further explanation of views, using VulkanHpp enums, and overall better restructuring

walcht:
This PR attempts to improve a couple of things in  the [Image Views](https://docs.vulkan.org/tutorial/latest/03_Drawing_a_triangle/01_Presentation/02_Image_views.html) chapter.

1. I think the explanation of the views can be enhanced by introducing a short explanation of buffer views which, I assume, are easier to explain than image views.

1. There is a mixture of C and C++ styles - Throughout this tutorial, C++ is used with VulkanHpp and in this chapter we have some code snippets like this:

   ```C++
   createInfo.components.r = VK_COMPONENT_SWIZZLE_IDENTITY;
   ```

1. There are some code snippets that are unnecessarily repeated (I think).
1. Finally, I did clang-format the code snippets so that they are within 80 column width limit - having to side scroll an embedded code snippet is not ideal.

Please do let me know if PRs like this are welcome (If so, I have done similar changes/improvements to other previous chapters that I can also open a PR for).

### DIFF
--- a/en/03_Drawing_a_triangle/01_Presentation/02_Image_views.adoc
+++ b/en/03_Drawing_a_triangle/01_Presentation/02_Image_views.adoc
@@ -2,11 +2,19 @@
 
 = Image views
 
-To use any `VkImage`, including those in the swap chain, in the render pipeline
-we have to create a `VkImageView` object. An image view is quite literally a
-view into an image. It describes how to access the image and which part of the
-image to access, for example, if it should be treated as a 2D texture depth
-texture without any mipmapping levels.
+In Vulkan, accessing and manipulating resources isn't done directly but rather
+through *views* - these describe how to look at (or *view*) a subset of the
+underlying data in a desired way. For instance, `VkBuffer` refers to a buffer
+object which represents a linear array of data. To use a `VkBuffer` you have to
+create a `VkBufferView` object which describes, among other things, the
+accessible contiguous range of the underlying data (this is done by providing
+an `offset` and a `range`).
+
+Similarly, to use any `VkImage`, including those in the swap chain, in the
+render pipeline we have to create a `VkImageView` object. The `VkImageView`
+describes how to interpret the underlying image and which part of it to access,
+for example, if it should be treated as a 2D texture depth texture without any
+mipmapping levels.
 
 In this chapter we'll write a `createImageViews` function that creates a basic
 image view for every image in the swap chain so that we can use them as color
@@ -20,7 +28,7 @@ std::vector<vk::raii::ImageView> swapChainImageViews;
 ----
 
 Create the `createImageViews` function and call it right after swap chain
-creation.
+creation:
 
 [,c++]
 ----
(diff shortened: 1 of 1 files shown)


SaschaWillems:
> There is a mixture of C and C++ styles - Throughout this tutorial, C++ is used with VulkanHpp and in this chapter we have some code snippets like this:

This is something we are going to fix, see #122. But removing the C syntax blocks might not be the right option, see #202.

> Finally, I did clang-format the code snippets so that they are within 80 column width limit - having to side scroll an embedded code snippet is not ideal.

For clang format: We plan on adding that, see #131. Once that's merged, we'll also try to adjust the documentation accordingly. The limit of 80 columns is something we need to discuss first. That was fine in the early days of programming, but nowadays might be too small.

But in general, for designated initializers, I prefer having values line-by-line, which improves readability a lot.

walcht:
Apologies if the PR seems to include a lot of things (this is my first in this project and I really want to contribute to it :-)).

I wasn't aware of #202 - it makes sense then to fix these language consistencies there.

Concerning #131, it is important to note that I am talking about embedded C++ formatting and not sample/attachment formatting. I don't know if a single clang-format file is planned for both or if they are separate. My reasoning for the 80 column width (or something like a 100 - definitely not a 160) is the following: some people might open an editor to the side and re-write the code snippets along the tutorial (that's at least me) - in this example, the longest line in the code snippet is 94 characters but is still occluded:

Screenshot from 2025-10-18 21-52-43

But this is just nitpicking and getting at least an agreed upon clang-format is good enough.

I will close this PR since most of the issues are already being addressed in other PRs.

gpx1000:
Thanks for the effort to help.  If you're willing it night make sense to combine efforts rather than ask you to simply close the pr.  You're more than welcome to target my branch so we can capture your work.  I think we need to add a contributor list as well and make things easier for those willing to help out.

walcht:
@gpx1000 - Then I will add the C snippets (the same way you are doing in the #202 PR) and open a PR targeting:
[gpx1000:use-multi-language-in-source](https://github.com/gpx1000/Vulkan-Tutorial/tree/use-multi-language-in-source). I will close this PR since I can't re-target to a different repository.

I want to help with adjusting other chapters (adding C snippets, fixing C++ inconsistencies, etc) - for instance I can address the [Drawing a triangle](https://docs.vulkan.org/tutorial/latest/03_Drawing_a_triangle/00_Setup/00_Base_code.html) > Setup sub-chapter. If you are ok with this, then you can just tell me which chapters I can help with so. Thanks.

gpx1000:
Sounds great!  Maybe post a list of which chapters you'd like to take on here and they will be all yours!  We will readjust if needed as we go.  Thanks for the help.

walcht:
Sure - I will start with:

-  **03_Drawing_a_triangle => 00_Setup/\***
- **03_Drawing_a_triangle => 02_Graphics_pipeline_basics/\***
- **03_Drawing_a_triangle => 01_Presentation => 01_Swap_chain.adoc**
- **03_Drawing_a_triangle => 01_Presentation => 02_Image_views.adoc** (done - see: [this PR](https://github.com/gpx1000/Vulkan-Tutorial/pull/1) )

I will commit each chapter (i.e., each adoc) file separately and open PRs targeting your branch [gpx1000:use-multi-language-in-source](https://github.com/gpx1000/Vulkan-Tutorial/tree/use-multi-language-in-source) (each PR targeting a single adoc chapter/file).

And, finally, just to confirm, I saw that the C language snippets are using C++ concepts (foreach loop, etc.) - I guess what is meant by C is that the Vulkan C-API will be used directly within C++ and not a complete re-write in C, right? (will assume it is just C Vulkan API).

Since I currently have some free time, I think I can finish this by tomorrow (Monday).

More than happy to contribute :-)

gpx1000:
Sounds great on all of that.  

And yes Vulkan c-api over cpp is the standard way we try to document rather than entirely c.

Issue #65, opened 2025-05-25
Semaphore setup may need rework


SaschaWillems:
The latest SDK added a new [VUID](https://docs.vulkan.org/spec/latest/chapters/cmdbuffers.html#VUID-vkQueueSubmit-pSignalSemaphores-00067) regarding the use of semaphores.

The tutorial seems to trigger this and we prob. need to create one semaphore per swapchain image instead of one per frame-in-flight (which we currently do).

SaschaWillems:
This has been mentioned in a recent thread on reddit. Looks like we need to fix/adjust the tutorial accordingly.

SaschaWillems:
Has been fixed with recent updates.

PR #207, opened 2025-10-22
Comment usage of waitIdle in chapter 15

asuessenbach:
You definitely don't need to `device.waitIdle()` after every frame. It's enough to wait in `cleanup()` (Well, with caution, see [VK_EXT_swapchain_maintenance1 extension](https://docs.vulkan.org/guide/latest/swapchain_semaphore_reuse.html#_vk_ext_swapchain_maintenance1_extension)).

The `queue.waitIdle()` in `drawFrame()` is not needed, when you have `swapChainImages.size()` `renderFinishedSemaphores`.

If this is considered to be ok, I can carry that change to all the other chapters.

### DIFF
--- a/attachments/15_hello_triangle.cpp
+++ b/attachments/15_hello_triangle.cpp
@@ -63,7 +63,7 @@ private:
     vk::raii::CommandBuffer commandBuffer = nullptr;
 
     vk::raii::Semaphore presentCompleteSemaphore = nullptr;
-    std::vector<vk::raii::Semaphore> renderFinishedSemaphores;
+    vk::raii::Semaphore renderFinishedSemaphore = nullptr;
     vk::raii::Fence drawFence = nullptr;
 
     std::vector<const char*> requiredDeviceExtension = {
@@ -106,6 +106,7 @@ private:
 
     void cleanup() {
         glfwDestroyWindow(window);
+
         glfwTerminate();
     }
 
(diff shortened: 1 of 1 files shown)


gpx1000:
Yep, definitely a good change.

asuessenbach:
> If this is considered to be ok, I can carry that change to all the other chapters.

I just realized, that none of the other chapters use `waitIdle` per frame, and all the other chapters are already using multiple `renderFinishedSemaphores`.
That is, this PR is ready for merging.

SaschaWillems:
Prob. something that's generally wrong in the tutorial: I think we also need to have a vector for the presentCompleteSemaphore semaphores that's sized by max. no of frames in flight.

See #65

charles-lunarg:
IIRC the wait idle is because frames in flight hasn't been introduced yet.

So it is there to make sure everything behave correctly with only one semaphore for acquire and one for present.

asuessenbach:
> IIRC the wait idle is because frames in flight hasn't been introduced yet.

That sounds reasonable!
But it should contain some very visible comment in the code then, that that `waitIdle` is just for this sample and should not be used in production code.

SaschaWillems:
The waitIdle is mentioned in the documentation side of the tutorial. But yes, that mention should at least add some detail on that and tell readers that waididle shouldn't be used.

charles-lunarg:
Agreed, waitIdle should have been better described.

I would be 100% for replacing waitIdle with proper usage of semaphores, especially the present semaphore array with size of swapchain image count. In fact, since it is best practice to have swapchain image count present semaphores, it would be better to only ever introduce present semaphores as an array (never as a single object) *because* the usage of the semaphore escapes outside of the applications direct control. 

I commented initially just to give context why, but I do not want to claim that it is the only way things must be done :smile:

SaschaWillems:
If we agree on this, documentation also needs to be updated to reflect this change.

gpx1000:
I 100% agree!  Will need to update documentation.  Sorry didn't have internet over the weekend; just now catching up.

asuessenbach:
I reverted most of my changes, as the use of multiple semaphores is introduced in the next chapter.
That is, it's just some comments added now.

gpx1000 (approved):

PR #219, opened 2025-11-05
Fix vector size for presentCompleteSemaphores.

asuessenbach:
Only the semaphores signalled on finished rendering need to be indexed by the image index. All other resources (command buffers, fences, semaphores signalled on present completion) are indexed by the frame index.

Includes some minor clean-up on chapter 15 as well.

As usual, if this is considered to be ok, I can carry this change to all the other chapters.

### DIFF
--- a/en/03_Drawing_a_triangle/00_Setup/01_Instance.adoc
+++ b/en/03_Drawing_a_triangle/00_Setup/01_Instance.adoc
@@ -150,14 +150,17 @@ Or use the tuple:
 
 [,c++]
 ----
-    auto [result, imageIndex] = swapChain->acquireNextImage( UINT64_MAX, **presentCompleteSemaphore[currentFrame], nullptr );
-    if (result == vk::Result::eErrorOutOfDateKHR) {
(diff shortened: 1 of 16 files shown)


SaschaWillems:
> As usual, if this is considered to be ok, I can carry this change to all the other chapters.

Definitively yes.

Without this change, the tutorial is not spec compliant, see #65

asuessenbach:
> but tutorial text needs to be adjusted to

If you could give me a hint, where the corresponding tutorial text is, I could try to adjust it accordingly.

SaschaWillems:
E.g. "03_Frames_in_flight.adoc". Looking at this, it does at least partially differ from the updated code and might need to be updated.

asuessenbach:
@SaschaWillems Ok, I've adjusted some documentation. Please have a closer look at it!

asuessenbach:
Anything else I'd need to adjust?

SaschaWillems (approved):

gpx1000 (approved):

PR #234, opened 2025-11-25
Fix the mipmap lesson: maxLod must be set in SamplerCreateInfo in order for mipmaps to work

slavc:
Fix .maxLod not being set in the SamplerCreateInfo, which leads to it being initialized to zero, which leads to mip mapping effectively being disabled with only the zeroth mip level being used.


slavc:
Relevant Vulkan specification documentation: https://docs.vulkan.org/spec/latest/chapters/samplers.html#VkSamplerCreateInfo .

SaschaWillems:
That's an embarrassing oversight. Thanks for fixing this.

Issue #121, opened 2025-08-04
Rendering and presentation chapter is partly confusing


SaschaWillems:
We need to work on the Drawing a triangle / Drawing / Rendering and presentation chapter. It's causing some confusion among readrs, esp. in the way it does "sync" via doing a queue wait idle. We shouldn't be doing that in a tutorial, esp. as the next chapter does sync in a proper way. That chapter also has lots of errors and doesn't match the actual code :/

gpx1000:
It looks like #246 fixed this?  closing as it looks right.  please reopen if you have things that are missed.

PR #246, opened 2025-12-07
Fix waitForFences usage and documentation

### LINKED ISSUE #124: Odd formatting for while loops
We have a lot of places where while loops are written like this:

```cpp
        while ( vk::Result::eTimeout == device.waitForFences( *inFlightFences[currentFrame], vk::True, UINT64_MAX ) )
            ;
```

This seems to be a short-hand version of while, hich is okay, but what looks odd is the line break and identation for the semicolon.

@gpx1000: Any reason the semicolon is on the next line and indented like that?

I'd prefer to clean up like this:

```cpp
        while ( vk::Result::eTimeout == device.waitForFences( *inFlightFences[currentFrame], vk::True, UINT64_MAX ) );
```

Or maybe even remove the while? We already wait for a very long time (UINT64_MAX), the while loop would prob. never do anything as the TDR (at least on Windows) already would've reset the driver.

### LINKED ISSUE #244: Result codes not always checked
While working on the tutorial I noticed that result codes aren't always checked. All samples (that present) use e.g. this line of code:

```cpp
auto [result, imageIndex] = swapChain.acquireNextImage(UINT64_MAX, *presentCompleteSemaphores[frameIndex], nullptr);`
```

But `result` is never checked and overwritten by another call a few lines later, so errors in this line aren't catched in a graceful manner.

A tutorial should teach proper error handling, so we should add proper error handling.

SaschaWillems:
This PR fixes the odd loop-formatting used for fence waits. It replaces it with proper result code checking, which the tutorial didn't do before. So timeouts for fences are now properly handled.

Also fixes several documentation related issues, like a errors and code differences in the presentation chapter and the code in the multi threading chapter not matching the actual code. 

Refs #121
Fixes #124 
Fixes #244

### DIFF
--- a/attachments/17_swap_chain_recreation.cpp
+++ b/attachments/17_swap_chain_recreation.cpp
@@ -524,7 +524,6 @@ class HelloTriangleApplication
 			throw std::runtime_error("failed to acquire swap chain image!");
 		}
 
-		device.resetFences(*inFlightFences[frameIndex]);
 		commandBuffers[frameIndex].reset();
 		recordCommandBuffer(imageIndex);
 


SaschaWillems:
@gpx1000: What should we do with mixing of C++ and C-Syntax? When fixing documentation, or adding stuff, I try to stay as close to the code as possible and use C++-Syntax in the documentation. That leads to a not-so-nice mix of vulkan-hpp and vulkan-h terms and names in some of the chapters. Shouldn't we just update all of them to vulkan-hpp instead?

en/03_Drawing_a_triangle/03_Drawing/02_Rendering_and_presentation.adoc
```diff
@@ -287,18 +301,10 @@ In our case we're using the `renderFinishedSemaphore` for that purpose.
 graphicsQueue.submit(submitInfo, *drawFence);
 ----
 
-We can now submit the command buffer to the graphics queue using `vkQueueSubmit`.
-The function takes an array of `VkSubmitInfo` structures as argument for efficiency when the workload is much larger.
+We can now submit the command buffer to the graphics queue using `submit`.
```
asuessenbach, line 304: Maybe, instead of just `submit`: `vk::Queue::submit`?

SaschaWillems: See my question above. I'm unsure myself as to what terminology to use. The tutorial is not very consistent in that regards.

PR #266, opened 2025-12-30
Added KHRCreateRenderpass2ExtensionName to requiredDeviceExtension to…

MortenKristiansen:
This fixes a warning from the validation layer when the shaders are added. The warning was:

```
Validation Error: [ VUID-VkShaderModuleCreateInfo-pCode-08740 ] | MessageID = 0x6e224e9
vkCreateShaderModule(): SPIR-V Capability DrawParameters was declared, but one of the following requirements is required (VkPhysicalDeviceVulkan11Features::shaderDrawParameters OR VK_KHR_shader_draw_parameters).
The Vulkan spec states: If pCode is a pointer to SPIR-V code, and pCode declares any of the capabilities listed in the SPIR-V Environment appendix, one of the corresponding requirements must be satisfied (https://docs.vulkan.org/spec/latest/chapters/shaders.html#VUID-VkShaderModuleCreateInfo-pCode-08740)

```

I've added the extension to all example code from that step forward as I expect it'll be required in them all.

### DIFF
--- a/attachments/09_shader_modules.cpp
+++ b/attachments/09_shader_modules.cpp
@@ -60,7 +60,6 @@ class HelloTriangleApplication
 	    vk::KHRSwapchainExtensionName,
 	    vk::KHRSpirv14ExtensionName,
 	    vk::KHRSynchronization2ExtensionName,
-	    vk::KHRCreateRenderpass2ExtensionName,
 	    vk::KHRShaderDrawParametersExtensionName};
 
 	void initWindow()
--- a/attachments/03_physical_device_selection.cpp
+++ b/attachments/03_physical_device_selection.cpp
@@ -50,8 +50,7 @@ class HelloTriangleApplication
 	std::vector<const char *> requiredDeviceExtension = {
 	    vk::KHRSwapchainExtensionName,
 	    vk::KHRSpirv14ExtensionName,
-	    vk::KHRSynchronization2ExtensionName,
-	    vk::KHRCreateRenderpass2ExtensionName};
+	    vk::KHRSynchronization2ExtensionName};
 
 	void initWindow()
 	{
--- a/attachments/04_logical_device.cpp
+++ b/attachments/04_logical_device.cpp
@@ -54,8 +54,7 @@ class HelloTriangleApplication
 	std::vector<const char *> requiredDeviceExtension = {
 	    vk::KHRSwapchainExtensionName,
 	    vk::KHRSpirv14ExtensionName,
-	    vk::KHRSynchronization2ExtensionName,
-	    vk::KHRCreateRenderpass2ExtensionName};
+	    vk::KHRSynchronization2ExtensionName};
 
 	void initWindow()
 	{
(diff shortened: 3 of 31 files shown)


en/03_Drawing_a_triangle/02_Graphics_pipeline_basics/01_Shader_modules.adoc
```diff
@@ -520,13 +520,24 @@ Modifying the structure to suit the fragment shader is easy:
...
 
-Finish by defining an array that contains these two structs, which we'll later use to reference them in the actual pipeline creation step.
+Defining an array that contains these two structs, which we'll later use to reference them in the actual pipeline creation step.
 
 [,c++]
 ----
 vk::PipelineShaderStageCreateInfo shaderStages[] = {vertShaderStageInfo, fragShaderStageInfo};
 ----
 
+Finish by adding DrawParameters to the deviceExtensions. Otherwise the
```
SaschaWillems, line 530: That's not a good explanation. Can you rephrase this so people understand why this is required when using Slang?

attachments/09_shader_modules.cpp
```diff
@@ -60,7 +60,8 @@ class HelloTriangleApplication
 	    vk::KHRSwapchainExtensionName,
 	    vk::KHRSpirv14ExtensionName,
 	    vk::KHRSynchronization2ExtensionName,
-	    vk::KHRCreateRenderpass2ExtensionName};
+	    vk::KHRCreateRenderpass2ExtensionName,
```
SaschaWillems: Only `VK_KHR_shader_draw_parameters` should be required. I don't think we need renderpass2 as that's core since 1.2 and we use 1.4.

SaschaWillems (changes requested):

attachments/09_shader_modules.cpp
MortenKristiansen: I can remove it, however it was there originally. I'll test if the validation layer complains.

en/03_Drawing_a_triangle/02_Graphics_pipeline_basics/01_Shader_modules.adoc
MortenKristiansen: Barely :-) I've just started learning Vulkan - working through the tutorial, fixing things here and there to get it working. 
As far as I can understand the SPIR compiler sets a flag or something in the bytecode signaling it's using parameters (though I don't see any at the moment). This cause the validation layer to complain the vk::KHRShaderDrawParametersExtensionName must be added.

What I can offer is to do some more research and come up with a better explanation. However, I can't make it tonight.

attachments/09_shader_modules.cpp
SaschaWillems: Sry my bad. If it was there, just leave it in place. I'm addressing extensions in #260

en/03_Drawing_a_triangle/02_Graphics_pipeline_basics/01_Shader_modules.adoc
SaschaWillems: Might be better to open issues for stuff like that instead. We'll then take a look and add a proper explanation.

MortenKristiansen: OK, I'll cancel this PR.

PR #279, opened 2026-01-13
Enable synchronization2 in feature chain

Raffaello:
otherwise an error is reported

### DIFF
--- a/en/03_Drawing_a_triangle/00_Setup/04_Logical_device_and_queues.adoc
+++ b/en/03_Drawing_a_triangle/00_Setup/04_Logical_device_and_queues.adoc
@@ -95,7 +95,9 @@ The C++ Vulkan API provides a helper template called `vk::StructureChain` that m
 // Create a chain of feature structures
 vk::StructureChain<vk::PhysicalDeviceFeatures2, vk::PhysicalDeviceVulkan13Features, vk::PhysicalDeviceExtendedDynamicStateFeaturesEXT> featureChain = {
     {},                               // vk::PhysicalDeviceFeatures2 (empty for now)
-    {.dynamicRendering = true },      // Enable dynamic rendering from Vulkan 1.3
+    {
+      .synchronization2 = true,
+      .dynamicRendering = true },      // Enable dynamic rendering from Vulkan 1.3
     {.extendedDynamicState = true }   // Enable extended dynamic state from the extension
 };
 ----


SaschaWillems (changes requested):
That chapter does not make use of any sycn2 features, so no need to enable that yet.

This should match what the code does instead:

```cpp
vk::StructureChain<vk::PhysicalDeviceFeatures2, vk::PhysicalDeviceVulkan13Features, vk::PhysicalDeviceExtendedDynamicStateFeaturesEXT> featureChain = {
		    {},                                   // vk::PhysicalDeviceFeatures2
		    {.dynamicRendering = true},           // vk::PhysicalDeviceVulkan13Features
		    {.extendedDynamicState = true}        // vk::PhysicalDeviceExtendedDynamicStateFeaturesEXT
		};
```

Raffaello:
following the tutorial, the program failed without sync2.
as it was explicitly wrote in the error message to enable it.
unless it was a bug in the validation layer.

besides i was using vulkan api 1.4 not 1.3 if it means anything.

p.s.
also there are quite few holes in the tutorial that i filled the gaps with old tutorial, it might also this one.

Raffaello:
besides please also note that in the C++ code example: https://docs.vulkan.org/tutorial/latest/_attachments/04_logical_device.cpp

has these required extensions:
```cpp
std::vector<const char *> requiredDeviceExtension = {
	    vk::KHRSwapchainExtensionName,
	    vk::KHRSpirv14ExtensionName,
	    vk::KHRSynchronization2ExtensionName,
	    vk::KHRCreateRenderpass2ExtensionName};
```

SaschaWillems:
We have recently removed most of those extensions, see #260. Docs.vulkan.org prob. hasn't been updated yet.

Raffaello:
> We have recently removed most of those extensions, see #260. Docs.vulkan.org prob. hasn't been updated yet.


@SaschaWillems thanks for that.
Yes it isn't. (http://docs.vulkan.org/tutorial/latest/03_Drawing_a_triangle/00_Setup/04_Logical_device_and_queues.html)

 I found also that in other pages is was referencing to the ` class QueueFamiliIndices` of the older tutorial but it was never mentioned in the new one as there is somewhere some `indices`
 ref: https://docs.vulkan.org/tutorial/latest/03_Drawing_a_triangle/01_Presentation/00_Window_surface.html
 
 like this page is "not ok", as well its code after: https://docs.vulkan.org/tutorial/latest/_attachments/05_window_surface.cpp
 doesn't neither have a word with `indices`. 
 I "understood" it, or better i am still try to grasp a better vulkan understanding as a beginner, becase i had started before with the older one: https://vulkan-tutorial.com/Drawing_a_triangle/Presentation/Window_surface
 
 and closed that gap. 
 but anyway it would be very great if the modern tutorial would be polished as sometimes is also not so linear to follow, something to understand will be understood a page after and requires to go back one page and so on. (but that might be ok as it is part of the learning process).

SaschaWillems:
The docs site is updated when spec updates are released, so updates can take up to a week before they show up there.

We are aware of the issue regarding differences in documentation and code, and they'll be fixed (see the other issues).

Fyi @gpx1000

Raffaello:
thanks

SaschaWillems:
The initial issue that this PR wanted to address has been resolved, so closing this.

PR #287, opened 2026-01-20
Remove attachment 18_vertex_input

### LINKED ISSUE #286: VVL warning in 18_vertex_input
I'm getting lots of VVL messages on running 18_vertex_input:
```
validation layer: type { Validation } msg: vkCmdDraw(): the last bound pipeline has pVertexBindingDescriptions[0].binding (0) (pointing to Locations [0, 1]) which didn't have a buffer bound from any vkCmdBindVertexBuffers call in this command buffer.
```
I think, we should either
- document (via output in the program!) that those messages are expected and resolved with 19_vertex_input 
  -> ok-ish, but might look a bit strange
- use the vertexInputInfo in createGraphicsPipeline and the actual shaders from the previous chapters
  -> the difference to 17_swap_chain_recreation would be almost negligible: just the struct Vertex and the array of vertices that are still not used
- just completely remove 18_vertex_input
  -> I'd vote for this one. What would you lose if this program were skipped?

asuessenbach:
Resolves #286.

### DIFF
--- a/en/04_Vertex_buffers/00_Vertex_input_description.adoc
+++ b/en/04_Vertex_buffers/00_Vertex_input_description.adoc
@@ -58,7 +58,8 @@ Create a new structure called `Vertex` with the two attributes that we're going 
 
 [,c++]
 ----
-struct Vertex {
+struct Vertex
+{
     glm::vec2 pos;
     glm::vec3 color;
 };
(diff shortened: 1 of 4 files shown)


asuessenbach:
> This requires documentation changes.

Done.

Issue #290, opened 2026-01-22
32_ecosystem_utilities fires a validation error, followed by a driver crash


gary-sweet:
I see:

```
Dynamic rendering supported via Vulkan 1.3
Timeline semaphores supported via Vulkan 1.2
Synchronization2 supported via Vulkan 1.3
Configuring pipeline for dynamic rendering
Creating timeline semaphores

Feature support summary:
- Dynamic Rendering: Yes
- Timeline Semaphores: Yes
- Synchronization2: Yes

validation layer: type { Validation } msg: vkQueueSubmit(): pSubmits[0].pSignalSemaphores[1] (VkSemaphore 0x2a000000002a) is a timeline semaphore, pSubmits[0].pNext<VkTimelineSemaphoreSubmitInfo>.signalSemaphoreValueCount (1) is different than pSubmits[0].signalSemaphoreCount (2).
The Vulkan spec states: If the pNext chain of this structure includes a VkTimelineSemaphoreSubmitInfo structure and any element of pSignalSemaphores was created with a VkSemaphoreType of VK_SEMAPHORE_TYPE_TIMELINE, then its signalSemaphoreValueCount member must equal signalSemaphoreCount (https://docs.vulkan.org/spec/latest/chapters/cmdbuffers.html#VUID-VkSubmitInfo-pNext-03241)
validation layer: type { Validation } msg: vkFreeCommandBuffers(): pCommandBuffers[0] (VkCommandBuffer 0x3ae89040) is in use.
The Vulkan spec states: All elements of pCommandBuffers must not be in the pending state (https://docs.vulkan.org/spec/latest/chapters/cmdbuffers.html#VUID-vkFreeCommandBuffers-pCommandBuffers-00047)
    02:01:30.871 nxserverlib: V3D platform(0x7f04002c70) unregistered
vk::raii::Queue::submit: ErrorDeviceLost
```

It looks like the semaphore counts don't correctly match.

PR #293, opened 2026-01-28
Fix timeline semaphore value counts

gpx1000:
to include binary semaphore values in submit info


gpx1000:
This should address #290

Issue #300, opened 2026-02-04
Skip recommendation of Vulkan-Hpp module


LordKobra:
Hello! I'm quite the beginner w.r.t. Vulkan. In 2024 i did the old Vulkan tutorial at https://vulkan-tutorial.com/
It mentioned being outdated and pointed me to this tutorial.

In the introduction, the new tutorial says https://docs.vulkan.org/tutorial/latest/00_Introduction.html
> Compared to the original tutorial, this version of the tutorial is teaching up-to-date concepts:
...
 Modern C++ (20) with modules

So modules are "sold" as one reason to follow through with this tutorial.

When we get to the development environment chapter it says

> Vulkan has support for C++ modules which became available with c++20. A large advantage of C++ modules is they give all the benefits of C++ without the overhead of long compile times. To do this, the .cppm file must be compiled for your target device. This tutorial demonstrates how to take advantage of C++ modules. However, to maximize compatibility across compilers and IDEs, the attachments template has C++20 module support disabled by default, but it is recommended to enable it if your toolchain supports it.
To enable the Vulkan C++20 module in the attachments template, configure CMake with: ...

So while it is disabled by default, I'm strongly encouraged to use modules. The problem is that module support is still very fractured and i had to spend many hours trying to get the module to compile. CMake forced me to update the project to C++23 standard. Upon asking the developers at Vulkan-Hpp two of them discouraged the usage of modules for beginners.
https://github.com/KhronosGroup/Vulkan-Hpp/issues/2463#issuecomment-3848577104
https://github.com/KhronosGroup/Vulkan-Hpp/issues/2463#issuecomment-3848620880
Edit: IntelliSense does not support modules either, so VSCode will throw many errors: https://github.com/microsoft/vscode-cpptools/issues/6302

In order to not confuse beginners like me, i recommend to discourage people from using modules, until involved parties (e.g. vulkan-hpp, cmake) update their support for modules and make it an overall better experience. Or maybe put the recommendation to an optional section at the end?

SaschaWillems:
See #310. We won't be removing them, as they'll soon become non-experimental.

Issue #299, opened 2026-02-04
Should the C++20 module be recommended to beginners?


M2-TE:
While the Vulkan-Hpp C++20 module can make Vulkan quite nice for beginners in the (hopefully near) future, it is currently still in a very experimental state and introduces breaking changes every now and then. This might frustrate newcomers who follow the tutorial.

Sections such as the one under `02_Development_environment` are already outdated:

https://github.com/KhronosGroup/Vulkan-Tutorial/blob/0aad08a13db459eff42718a47bd05ce8590a132a/en/02_Development_environment.adoc?plain=1#L127-L184

Once the module changes in KhronosGroup/Vulkan-Hpp#2443 are complete, I can update the tutorial sections regarding modules here, similar to the updates to our own README and tutorial we are working on in KhronosGroup/Vulkan-Hpp#2358.

Until then, it would probably be a good idea to very explictly mark that section as experimental, or at least write the matching SDK version for the shown sample.

LordKobra:
I see we both had the same idea. My issue is a duplicate https://github.com/KhronosGroup/Vulkan-Tutorial/issues/300
(I will keep it up for now though, because it's a different viewpoint: vulkan beginner)

SaschaWillems:
See #301. We won't be removing them, as they'll soon become non-experimental.

M2-TE:
I reckon you mean #301 ;)

The warning notice in that PR is good, I didn't want to make it sound like I wanted it removed.

We aren't really planning on making them non-experimental that soon though, since CMake support itself is still experimental. Once `import std` lands in one of their stable releases, we will make a stable Vulkan-Hpp modules release as well!

gpx1000:
I was hopeful to not have to make a more permanent change along the lines of completely removing even knowledge of Modules existing as they come with massive benefits to compile time.

However, if it's really going to be a long term horizon as to where it's not really expected to be a next few year thing, or somewhere in the reasonable time frame thing, then I don't think we're doing tutorial users any benefit by even mentioning they exist.  We can chalk it up to my mistake for putting them in to begin with and remove everything we can find about them from the tutorial.  
It is starting to seem like it'd be easier to just rework the tutorial to include them in a few years when they are available rather than give the impression they exist right now and have an easy to change note to discourage their use until they're available.  I guess it's something for us to discuss at the next docs call.

M2-TE:
[...]

PR #301, opened 2026-02-05
Update wording on modules

SaschaWillems:
With modules support in vulkan-hpp still experimental, we might want to adjust wording so people know that they shouldn't be used for now. This PR adjusts wording around modules.

Refs #300
Refs #299


gpx1000 (approved):

SaschaWillems:
As discussed today: We might want to slightly rephrase this and merge. Even with VS 2026 at least Intellisense-Support for Moduls is still reported as experimentl, so people should be aware of that.

gpx1000 (approved):

PR #319, opened 2026-02-27
Overview: Restoring Step 4 from "Dynamic Rendering" back to "framebuffers"

ohmaya:


SaschaWillems:
Not sure about this. We don't use frame buffers anymore with dynamic rendering. IMO we should rewrite that whole chapter to properly reflect what the tutorial does.

ohmaya:
@SaschaWillems 
See the revision history, once Step 4 had Dynamic Rendering content added and later the Dynamic Rendering content was moved to Step 5, now Step 4 basically has the same content (for image views and framebuffers) as Alexander Overvoorde's version, however, the title was not restored.

SaschaWillems:
We discussed this today and we'll be closing this. Restoring the caption makes no sense, as our version of the tutorial does not use framebuffers. So what I did instead was adjust the paragraph's text to reflect this. See #404.

Issue #354, opened 2026-04-09
22_descriptor_layout fails with VVL error and crash


asuessenbach:
In my environment (Win11, VS 2022), 22_descriptor_layout fails to run, calling the `debugCallback` with the following error:

```
vkCreateGraphicsPipelines(): pCreateInfos[0].pStages[0] SPIR-V (VK_SHADER_STAGE_VERTEX_BIT) uses descriptor [Set 0, Binding 0, variable "ubo"] but the binding was not declared in the VkPipelineLayoutCreateInfo::pSetLayouts[0].
Possible VkDescriptorType that could be used are: VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER_DYNAMIC or VK_DESCRIPTOR_TYPE_INLINE_UNIFORM_BLOCK or VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER.
The Vulkan spec states: If a resource variable is declared in a shader and layout is not VK_NULL_HANDLE, the corresponding descriptor set in layout must match the shader stage (https://docs.vulkan.org/spec/latest/chapters/pipelines.html#VUID-VkGraphicsPipelineCreateInfo-layout-07988)
```

Subsequently, it crashes with an `Access violation reading location 0x0000000000000028` in `drawIndexed`.

Is that the expected behaviour? The corresponding docu `00_Descriptor_set_layout_and_buffer.adoc` does not mention it.

Should that be fixed, even though it might need some parts introduced with 23_descriptor_sets?
Or should we just use the 19_shader_vertexbuffer again, instead of the 22_shader_ubo?

asuessenbach:
Any opinion on this issue, anyone?

Wingitalways:
Not sure if I'm allowed to respond here so sorry if not! But I had the same issue working through the tutorial and if you haven't found it already, you have to change the .setLayoutCount in the "pipelineLayoutInfo" to = 1 instead of = 0, and add the parameter ".pSetLayouts to = &*descriptorSetLayout", which should fix the issue!

For some reason the example code doesn't add a descriptor set count and layout, hopefully it can get changed soon!

PR #357, opened 2026-04-15
Align docu in 00_Descriptor_set_layout_and_buffer.adoc and sources in 22_descriptor_layout.cpp

asuessenbach:
Note that this sample does not run. It's failing due to missing binding of a descriptor set. Which is done in the next chapter.

Should we add a few words to `00_Descriptor_set_layout_and_buffer.adoc` to explain that failure, or should we just remove this chapter and add its content to the next one?

See also #354.

### DIFF
--- a/attachments/22_descriptor_layout.cpp
+++ b/attachments/22_descriptor_layout.cpp
@@ -44,14 +44,13 @@ struct Vertex
 
 	static vk::VertexInputBindingDescription getBindingDescription()
 	{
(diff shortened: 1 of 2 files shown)


gpx1000 (approved):

SaschaWillems:
I'm pretty sure that it's mentioned somewhere that not all chapters are intended to display something up until to a certain point. At least the original tutorial did.

I don't think removing things that don't work yet is something we should do, as that's just how the tutorial is/was structured.

SaschaWillems (approved):

PR #358, opened 2026-04-16
Align docu in 01_Descriptor_pool_and_sets.adoc and sources in 23_descriptor_sets.cpp

asuessenbach:
<del>Unfortunately, I just see a black window when running this sample.
If someone could figure out the error, I'ld be happy to fix it with this PR.</del>


SaschaWillems:
> If someone could figure out the error, I'ld be happy to fix it with this PR.

Issue is wrong winding order. Can be fixed by changing `frontFace` to `vk::FrontFace::eCounterClockwise`.

asuessenbach:
> Can be fixed by changing frontFace to vk::FrontFace::eCounterClockwise.

And that is explicitly mentioned that way in the doc file here!!

SaschaWillems (approved):

PR #361, opened 2026-04-28
Add CMake package/dependency find modules to template project

dorileo:
Avoid frustration for newcomers and offer a last resort for resolving dependencies. The dependencies install script may fail in certain OSes or distributions, adding the find modules to template project improves the experience once it falls back to FetchContent in case the package is not found.

This change also adopts the "CMake" dir name convention present everywhere in the code base (instead of the originally expected "cmake").


marty-johnson59:
Hi @dorileo, thanks for your suggestion.  The team reviewed this and decided not to proceed at the moment as we believe we already have clear instructions in place describing how this works.  If you feel there are specific issues that aren't being addressed, please open an issue on this repo and we'll take a look.  Thanks!

PR #374, opened 2026-05-14
chore: Add REUSE compliance and AI block

KhronosWebservices:
- add reuse compliance to the repository
- add an AI Contributing block to the CONTRIBUTING doc
- add an Open Source Apache 2.0  CLA to cover in-bound contributions.  

This repository CONTRIBUTING.adoc originally listed only CC-BY-SA-4.0. However, the repository also has Apache-2.0, CC-BY-4.0 and  BSD-3-Clause.

As there are typically a lot of questions around the use of wildcards in REUSE.toml, here is a little background:

The wildcard has been put in place solely because REUSE was introduced without first updating every file. A strict configuration would have blocked any commit touching a non-compliant file, which probably would not make the maintainers very happy. The wildcard gives us a compliant baseline while headers are added at the maintainers' pace.

REUSE annotations are not just for exceptions. The REUSE spec fully supports REUSE.toml as the primary mechanism for declaring copyright and licensing. File headers are recommended, not required.

1. Wildcards can be removed once file headers are in place across the repository.
2. Updating headers is the correct thing to do. However, the wildcard is a valid way to use REUSE.toml.
3. Once the wildcard is removed, missing licences will trigger lint failures.

### DIFF
--- a/REUSE.toml
+++ b/REUSE.toml
@@ -1,0 +1,47 @@
+version = 1
+
+[[annotations]]
+path = [
+    "**.adoc",
+    "images/**",
+]
+SPDX-FileCopyrightText = "The Khronos Group, Inc."
+SPDX-License-Identifier = "CC-BY-4.0"
+
+[[annotations]]
+path ="attachments/**"
+SPDX-FileCopyrightText = "The Khronos Group, Inc."
+SPDX-License-Identifier = "Apache-2.0"
+
+[[annotations]]
+path = [
+    "antora-ci-playbook.yml",
+    "antora/**",
+    ".github/workflows/**",
+]
+precedence = "closest"
+SPDX-FileCopyrightText = "Sascha Willems"
+SPDX-License-Identifier = "CC-BY-4.0"
+
+[[annotations]]
+path = [
+    ".github/workflows/reuse-compliance.yml",
+    ".gitignore",
+]
+precedence = "closest"
+SPDX-FileCopyrightText = "The Khronos Group, Inc."
+SPDX-License-Identifier = "CC-BY-4.0"
+
+[[annotations]]
+path = "attachments/CMake/FindVulkan.cmake"
+SPDX-FileCopyrightText = "Holochip Inc."
+SPDX-License-Identifier = "BSD-3-Clause"
+
+[[annotations]]
+path = [
+    "attachments/simple_engine/**",
+    "scripts/**",
+]
+SPDX-FileCopyrightText = "Holochip Corporation"
+SPDX-License-Identifier = "Apache-2.0"
+
--- a/REUSE.toml
+++ b/REUSE.toml
@@ -148,6 +148,7 @@ precedence = "closest"
 SPDX-FileCopyrightText = [
     "2023-2026 The Khronos Group, Inc.",
 ]
+
 SPDX-License-Identifier = "Apache-2.0"
 
 [[annotations]]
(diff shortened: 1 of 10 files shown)


SaschaWillems:
Looks good to me.

I do have one question though: [This](https://github.com/KhronosGroup/Vulkan-Tutorial/pull/374/changes#diff-74bd2d6afd7a1f967679adae374ca38fd9745fe6129dc4717a1ab0e388590b08R11) has Apache-2.0 as license. But that folder also contains the tutorial source, which is base on the original tutorial code that afaik is CC-BY SA.

Or does the REUSE only apply to files that contain a copyright header?

KhronosWebservices:
Yes, I did miss that in the Overview. I am thinking that not all files in `/en` should be copyright both Alexander Overvoorde and Khronos Group?

KhronosWebservices:
> Or does the REUSE only apply to files that contain a copyright header?

REUSE applies to all files included in the `path` statement under [[annotations]]. Using a wildcard /en/** would apply to all files in /en. Ideally we would want each file to have their own SDPX Copyright and License, but that is a big undertaking. REUSE.toml is a valid work around until each file  can be updated.

REUSE.toml
```diff
@@ -0,0 +1,57 @@
...
+    "images/**",
+]
+SPDX-FileCopyrightText = "The Khronos Group, Inc."
+SPDX-License-Identifier = "CC-BY-4.0"
+
+[[annotations]]
+path = [
+    "en/**",
+]
+SPDX-FileCopyrightText = [
```
SaschaWillems, line 56: Not sure about this. Our tutorial is based on https://github.com/Overv/VulkanTutorial, so shouldn't this somehow at least mention the author of the original Vulkan tutorial?

KhronosWebservices: Yes, that is a very good idea, thank you @SaschaWillems. Should the copyright be something along the lines of:

2023 Alexander Overvoorde
2023-2026 The Khronos Group Inc.

REUSE.toml
```diff
@@ -0,0 +1,57 @@
...
+    "The Khronos Group, Inc.",
+]
+SPDX-License-Identifier = "CC-BY-SA-4.0"
+
+
+[[annotations]]
+path ="attachments/**"
+SPDX-FileCopyrightText = "The Khronos Group, Inc."
+SPDX-License-Identifier = "Apache-2.0"
```
KhronosWebservices, line 76: Should this be CC-BY-1.0 to match the old /code directory from Overv?

REUSE.toml
gpx1000: Sorry, it's slightly more complicated.  Alexander's original code only covers up to 10_multisampling.adoc.  It shouldn't cover the rest of the tutorial.  I guess I should add the Holochip copyright, but I'm fine with just leaving it as Khronos group everywhere.

REUSE.toml
gpx1000: Only the top level base tutorial should be CC-BY-1.0.  The Simple Game Engine should be the more permissive Apache-2.0

KhronosWebservices:
REUSE.toml updated. If anyone wishes to do a file-by-file verification, running `reuse spdx` will create a SPDX bill of materials.

SaschaWillems (approved):
Following up on the Slack discussion I'm fine with merging this as-is.

KhronosWebservices:
@gpx1000 Any objection to merging?

gpx1000 (approved):

PR #392, opened 2026-06-08
Update the table of contents to reflect the addition of dynamic rendering

kareertl:
https://docs.vulkan.org/tutorial/latest/03_Drawing_a_triangle/02_Graphics_pipeline_basics/03_Render_passes.html was been updated with dynamic rendering info, but the table of contents still mentions only "Render passes." Update the table of contents to reflect the change.

### DIFF
--- a/antora/modules/ROOT/nav.adoc
+++ b/antora/modules/ROOT/nav.adoc
@@ -21,7 +21,7 @@
 *** xref:03_Drawing_a_triangle/02_Graphics_pipeline_basics/00_Introduction.adoc[Introduction]
 *** xref:03_Drawing_a_triangle/02_Graphics_pipeline_basics/01_Shader_modules.adoc[Shader modules]
 *** xref:03_Drawing_a_triangle/02_Graphics_pipeline_basics/02_Fixed_functions.adoc[Fixed functions]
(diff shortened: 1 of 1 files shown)


SaschaWillems:
They're still called render passes, no matter if you use the old style api or dynamic rendering.

kareertl:
> They're still called render passes, no matter if you use the old style api or dynamic rendering.

Yes, hence not removing "Render passes." The problem is the inconsistency of the table of contents and the updated chapter title on the page:

Render_passes

It'd be even better for the tutorial to compare the semantic meanings of the terms, "render pass," "traditional render pass," "VkRenderPass," "render pass object/struct," "dynamic rendering," "dynamic render pass," etc., elsewhere on the Internet (e.g. Reddit, some Q&A websites, etc.) and their real definitions to reduce beginner confusion.

Possible TODOs:

- Introduce the render pass evolution (?) history briefly to provide the context of the terms
- List the definitions of the terms
- Mention frequent incorrect/inaccurate usages of the terms
- Compare the definitions and the incorrect/inaccurate usages

SaschaWillems:
As for the naming confusion: We'll fix that with #408

As for the possible Todos: The docs site covers all of that. No need to duplicate that information here in the tutorial.

Issue #274, opened 2026-01-11
Multisampling chapter still uses render passes


rhythm1705:
Rest of the tutorial has moved to dynamic rendering, but this chapter still refers to render passes:

<img width="1151" height="442" alt="Image" src="https://github.com/user-attachments/assets/e8a9c51a-c701-41e1-b2c2-bb56dfbc4a4c" />

PR #418, opened 2026-07-09
Align docu in 10_Multisampling.adoc and 30_multisampling.cpp

asuessenbach:


SaschaWillems:
Not sure about this. The MSAA chapter (and code) are both outdated and still use Renderpasses. So we should first update these to dynamic rendering. See #274

asuessenbach:
> The MSAA chapter (and code) are both outdated and still use Renderpasses.

No, it does not. The code did not contain any RenderPass handling, and the docu now also doesn't mention that stuff.

SaschaWillems:
Looks like it has been mostly adjusted, without the issue being closed. But the documentation still e.g. has a code paragraph for subpasses :/

SaschaWillems (approved):

PR #428, opened 2026-07-16
Add in updated contributors and the LLM disclosure.

gpx1000:


marty-johnson59:
Merging to get this in before SIGGRAPH.  @iagoCL, feel free to tweak w/ new PR if the Arm stuff needs changes.  Thx

README.adoc
```diff
@@ -7,6 +7,22 @@
...
+* xref:ML_Inference/introduction.adoc[Machine learning inference]
+* xref:OpenXR_Vulkan_Spatial_Computing/introduction.adoc[Vulkan for XR]
+* xref:AI_Assisted_Vulkan/introduction.adoc[LLM-Assistant Vulkan Development]
+* xref:Advanced_glTF/introduction.adoc[Advanced glTF]
+* xref:Advanced_Vulkan_Compute/introduction.adoc[Advanced Vulkan Compute]
+* xref:Building_a_Simple_Engine/introduction.adoc[Building a Simple Game Engine]
+* xref:Synchronization/introduction.adoc[Synchronization]
+* xref:Building_a_Simple_Engine/Courses/Opacity_Micromaps/00_introduction.adoc[Opacity Micromaps mini course]
+
+The following courses were written by Arm Limited and released under link:https://creativecommons.org/licenses/by-sa/4.0/[CC BY-SA 4.0].
```
SaschaWillems, line 21: At least the SIGGRAPH 2026 one is licensed under [MIT](https://github.com/KhronosGroup/Vulkan-Tutorial/blob/main/attachments/siggraph2026_vk_tutorial/LICENSE)

Raytracing does not state what license it falls under.

iagoCL: We choose MIT for the code as it is a common and simple license, so code in tutorials using it are easy to adopt by developers, and MIT should be "compatible" with cc-by-4.0.

The default license for the repository is cc-by-4.0 so it should apply to the course content.
We can consider adding or changing the licenses  if it will make things easier.

gpx1000: As this PR was merged, I did remove the "and released under..." portion of the sentence as I didn't want to speak for Arm.  You guys can select whichever license.  We (Holochip) are already using a different license for the new tutorials those are wholly separate from the core tutorial.
The first course does depend upon the core tutorial so I thought it might want to chose to remain on the same CC-BY but again i'm not speaking for you guys ;-).  I think the new course is wholly separate so it makes sense to be in whichever license you want.

iagoCL: We added the MIT License for the code, but my understanding is that the course materials and slides are licensed under CC BY-SA 4.0.

This should allow people to reuse and adapt the code in their own projects while also adapting and extending the course materials, which we're happy to encourage. For example, universities could use it as the basis for their own courses, or people could create Vulkan courses in other languages.

We're happy to leave things as they are, but we can also make changes if you prefer. For example, we could:

- Revisit the licensing for our content, we can consider relicensing the Arm tutorials under MIT, Apache 2.0, CC BY 4.0, or an appropriate combination. I think the current licensing arrangement works well. SIGGRAPH content is already cc-by 4.0
- Add a comment to the relevant files explaining which license applies to each file. If we expect to continue using multiple licenses, it would probably be a good idea to start adding something like the following to all AsciiDoc files:

```asciiDoc
// Copyright (c) 2026 Khronos Group
// Copyright (c) 2026 Arm
// SPDX-License-Identifier: CC-BY-4.0 OR Apache-2.0
```
I would keep the licenses as they are, but personally I think we should require adding a comment with the license to all .adoc files in the repository.

PR #427, opened 2026-07-16
Remove unused helper function

SaschaWillems:
This PR removes a helper function not used (anymore).

Fixes #46

PR #440, opened 2026-07-29
Fix depth-stencil image aspect masks

swayam8624:
## Summary
- derive the depth image aspect mask from the selected depth format
- use the mask for the image view and both layout-transition barriers

## Root cause
The tutorial can select a depth/stencil format such as VK_FORMAT_D32_SFLOAT_S8_UINT, while the attachment sample hardcodes VK_IMAGE_ASPECT_DEPTH_BIT. Validation then reports VUID-VkImageMemoryBarrier2-image-03320 because the stencil aspect is omitted.

## Validation
- Reproduced on macOS Apple Silicon (M2 Pro) with the Khronos validation layer.
- `cmake -S attachments -B build/tutorial -G Ninja -DCMAKE_BUILD_TYPE=Debug -DCMAKE_OSX_ARCHITECTURES=arm64 -DENABLE_CPP20_MODULE=OFF -DKTX_INCLUDE_DIR=... -DKTX_LIBRARY=...`
- `cmake --build build/tutorial --target 32_ecosystem_utilities -j4`
- standalone C++20 syntax check passed with Vulkan-Hpp defines
- `git diff --check` passed

The executable initializes Vulkan but exits in this checkout because the required tutorial asset file is not found from the build layout.

### DIFF
--- a/attachments/32_ecosystem_utilities.cpp
+++ b/attachments/32_ecosystem_utilities.cpp
@@ -11,6 +11,7 @@
 #include <optional>
 #include <stdexcept>
 #include <vector>
+#include <vulkan/vulkan_format_traits.hpp>
 
 #if defined(__INTELLISENSE__) || !defined(USE_CPP20_MODULES)
 #	include <vulkan/vulkan_raii.hpp>
@@ -813,7 +814,7 @@ class HelloTriangleApplication
 	{
 		vk::Format depthFormat = findDepthFormat();
 		depthImageAspect       = vk::ImageAspectFlagBits::eDepth;
-		if (hasStencilComponent(depthFormat))
+		if (vk::hasStencilComponent(depthFormat))
 		{
 			depthImageAspect |= vk::ImageAspectFlagBits::eStencil;
 		}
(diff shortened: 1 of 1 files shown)


attachments/32_ecosystem_utilities.cpp
```diff
@@ -811,9 +812,14 @@ class HelloTriangleApplication
 	void createDepthResources()
 	{
 		vk::Format depthFormat = findDepthFormat();
+		depthImageAspect       = vk::ImageAspectFlagBits::eDepth;
+		if (hasStencilComponent(depthFormat))
```
asuessenbach: Instead of the potentially being removed `Renderer::hasStencilComponent` (see #427), you might use the more general `vk::hasStencilComponent` from `vulkan_format_traits.hpp`.

asuessenbach (commented):
An important fix.
Just one issue:

swayam8624:
Addressed in commit f2b8e25. I added the explicit `vulkan/vulkan_format_traits.hpp` include, replaced the attachment-local helper with `vk::hasStencilComponent`, and removed the duplicate helper. The `32_ecosystem_utilities` target builds successfully on macOS arm64.

asuessenbach (approved):

Issue #436, opened 2026-07-20
Validtion error in lesson 03.03.01 Command buffers


peekagrub:
Following the tutorial there is an validation error from using `vkCmdPipelineBarrier2()` without enabling the extension. This function was introduced for placing a barrier for [transitioning the image layout](https://github.com/KhronosGroup/Vulkan-Tutorial/blob/main/en/03_Drawing_a_triangle/03_Drawing/01_Command_buffers.adoc#image-layout-transitions) of the framebuffer. 
This function requires an extension to use but the tutorial doesn't mention this being required and the linked C++ code also doesn't update either the `isDeviceSuitable()` or `createLogicalDevice()` functions to reflect this new required extension.
The C++ code in the next chapter does have `isDeviceSuitable()` and `createLogicalDevice()` functions to enable this feature by checking and setting `VkPhysicalDeviceVulkan13Features.synchronization2`.
```cpp
bool supportsRequiredFeatures = features.template get<vk::PhysicalDeviceVulkan11Features>().shaderDrawParameters &&
                                features.template get<vk::PhysicalDeviceVulkan13Features>().dynamicRendering &&
                                features.template get<vk::PhysicalDeviceVulkan13Features>().synchronization2 &&
                                features.template get<vk::PhysicalDeviceExtendedDynamicStateFeaturesEXT>().extendedDynamicState;
```
```cpp
// query for Vulkan 1.3 features
vk::StructureChain<vk::PhysicalDeviceFeatures2,
                   vk::PhysicalDeviceVulkan11Features,
                   vk::PhysicalDeviceVulkan13Features,
                   vk::PhysicalDeviceExtendedDynamicStateFeaturesEXT>
    featureChain = {
        {},                                                          // vk::PhysicalDeviceFeatures2
        {.shaderDrawParameters = true},                              // vk::PhysicalDeviceVulkan11Features
        {.synchronization2 = true, .dynamicRendering = true},        // vk::PhysicalDeviceVulkan13Features
        {.extendedDynamicState = true}                               // vk::PhysicalDeviceExtendedDynamicStateFeaturesEXT
    };
```

PR #442, opened 2026-08-06
Add missing synchronization2 feature check and enable flag

CatB1t:
Neither the docs nor the source files check for or enable the synchronization2
feature before `15_hello_triangle.cpp`, which uses the sync2 API. It's correctly enabled
in that chapter's own source, but not in anything prior to it, so anyone following the
tutorial in order hits a validation error (feature not enabled) when they get there.

This PR adds the support check and the enable flag to the docs and the attachment
sources for the chapters leading up to it.

Solves #436


asuessenbach:
As far as I can tell, no synchronization2 feature is used before 15_hello_triangle.
The only thing that's missing, though, is some description on adding this flag to `createLogicalDevice` and `isDeviceSuitable` for that chapter (`02_Rendering_and presentation.adoc`).

CatB1t:
> As far as I can tell, no synchronization2 feature is used before 15_hello_triangle. The only thing that's missing, though, is some description on adding this flag to `createLogicalDevice` and `isDeviceSuitable` for that chapter (`02_Rendering_and presentation.adoc`).

I was trying to follow the pattern of enabling required features & extensions beforehand for the `Hello Triangle` demo, I'll add a dedicated section instead for that chapter.

PR #458, opened 2026-08-19
Fix Android CI.

gpx1000:
Fingers crossed!!!  This *should* be the fix.


SaschaWillems:
I does indeed work. Thank you very much 👍🏻

SaschaWillems (approved):

PR #461, opened 2026-08-19
Implement reference counting in ResourceManager (#320)

### LINKED ISSUE #320: ResourceManager doesn't implement Reference Counting.
According to [04_resource_management.adoc](https://github.com/KhronosGroup/Vulkan-Tutorial/blob/main/en/Building_a_Simple_Engine/Engine_Architecture/04_resource_management.adoc),

> When designing a resource management system, you’ll need to address several challenges:
> 3. Reference Counting - Track how many objects are using a resource to know when it can be safely unloaded.

This seems to be implemented as `int ResourceManager::ResourceData::refCount`, but it's never modified in [04_resource_management.adoc](https://github.com/KhronosGroup/Vulkan-Tutorial/blob/main/en/Building_a_Simple_Engine/Engine_Architecture/04_resource_management.adoc), and doesn't exist in [resource_manager.h](https://github.com/KhronosGroup/Vulkan-Tutorial/blob/main/attachments/simple_engine/resource_manager.h).

gpx1000:
refCount was documented as part of the resource management design but never implemented: LoadResource never incremented it and UnloadResource unloaded and erased a resource unconditionally, even with other outstanding handles to it.

Also drop the `final` on Resource and ResourceManager: as written, neither could ever be subclassed, so none of the concrete resource types the tutorial documents (Texture, Mesh, Shader) could actually exist.

Fixes #320

### DIFF
--- a/attachments/simple_engine/resource_manager.cpp
+++ b/attachments/simple_engine/resource_manager.cpp
@@ -43,7 +43,7 @@ void ResourceManager::UnloadAllResources()
 		for (auto &innerKv : val)
 		{
 			auto &loadedResource = innerKv.second;
-			loadedResource->Unload();
+			loadedResource.resource->Unload();
 		}
 		val.clear();
 	}
--- a/attachments/simple_engine/resource_manager.h
+++ b/attachments/simple_engine/resource_manager.h
@@ -26,7 +26,7 @@
 /**
  * @brief Base class for all resources.
  */
-class Resource final
+class Resource
 {
   protected:
 	std::string resourceId;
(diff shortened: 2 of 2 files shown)


asuessenbach:
Why do you implement your own ref counting, instead of just using a `std::shared_ptr`?

gpx1000:
Simply for the purposes of educational content showing specifically how ref counting works.  shared_ptr naturally is a better use so I can be convinced to reimplement as shared_ptr.

asuessenbach:
> Simply for the purposes of educational content showing specifically how ref counting works.

I don't think, ref counting needs to be demonstrated here. It's not that vulkan specific.
That is: I'ld vote for using `std::shared_ptr`.

gpx1000:
switched to std::shared_ptr

SaschaWillems (approved):

SaschaWillems:
There's also #504, which seems to adjust the documentation, whereas this only adjusts code. Could that be done in a single PR?

PR #469, opened 2026-08-20
Fix SV_Target case mismatch and stale prose in shader modules chapter

### LINKED ISSUE #112: Lesson 9, I like to whine
This bit of text is odd:

> The built-in annotation SV_Position functions as the output. Within the VertexOutput struct. Something worth mentioning if you’re familiar with other shading languages like GLSL or HLSL, there are no instructions for bindings.

Is this three sentences or two?

Not an error but a suggestion: stick with Slang throughout as far as compilation goes (you can keep the GLSL files as examples). Trying to reverse-engineer the existing CMakeLists.txt in the `attachments` directory for one's own code is complicated by trying to account for both.

Text says `SV_TARGET`, code says `SV_Target`.

There's a mention of `location` directives, which seems to be a remnant of the original vulkan-tutorial.com version of this lesson that used GLSL fragment shaders.

Also not an error, but there is a paragraph starting with "At the time of writing..." justifying the use of SPIR-V 1.4, which wikipedia says was released 5 years ago.

In this code block:

```
add_slang_shader_target( foo SOURCES ${SHADER_SLANG_SOURCES})
target_add_dependencies(bar PUBLIC foo)
```

It's not clear to me what the second line means. And should it be `target_add_dependencies` or `add_dependencies` which is closer to what's in the sample code's CMakeLists.txt?

Anyway, I had trouble getting the code to compile using the `add_slang_shader_target` function in the tutorial until I added a dummy variable and passed a dummy string to it, because cmake was complaining that I'd defined the keyword "SOURCE" more than once.  Apparently `cmake_parse_arguments` needs at least four arguments and I think if you pass it fewer then it just doubles one of the parameters?  The Cmake documentation says up front that the implementation of `cmake_parse_arguments` is "naive".  Anyway, I had to do this:

```
function(add_slang_shader_target TARGET)
  cmake_parse_arguments("SHADER" "" "SOURCES" "POINTLESS" ${ARGN})
  set(SHADERS_DIR ${CMAKE_CURRENT_BINARY_DIR}/shader)
  set(ENTRY_POINTS -entry vertMain -entry fragMain)
  add_custom_command(
    OUTPUT ${SHADERS_DIR}
    COMMAND ${CMAKE_COMMAND} -E make_directory ${SHADERS_DIR}
  )
  add_custom_command(
    OUTPUT ${SHADERS_DIR}/slang.spv
    COMMAND ${SLANGC_EXECUTABLE} ${SHADER_SOURCES} -target spirv -profile spirv_1_4 -emit-spirv-directly -fvk-use-entrypoint-name ${ENTRY_POINTS} -o slang.spv
    WORKING_DIRECTORY ${SHADERS_DIR}
    DEPENDS ${SHADERS_DIR} ${SHADER_SOURCES}
    COMMENT "Compiling Slang Shaders"
    VERBATIM
  )
  add_custom_target(${TARGET} DEPENDS ${SHADERS_DIR}/slang.spv)
endfunction()
```

And then this:

```
add_slang_shader_target(${SLANG_SHADER_TARGET} SOURCES ${SHADER_SOURCE_FILES} POINTLESS "FOO")
add_dependencies(${PROJECT_NAME} ${SLANG_SHADER_TARGET})
```

And that worked.

In the function `readFile` if you're trying to keep the C++20 thing going then you might switch from raw string paths to std::filesystem

More C API references in multiple places (e.g. `VkFoo` instead of `vk::foo`).

gpx1000:
Fixes the `SV_TARGET`/`SV_Target` case mismatch and its inaccurate description of a nonexistent `outColor` variable, and removes an outdated "at the time of writing" hedge about SPIR-V 1.4 support.

Fixes #112.

### DIFF
--- a/en/03_Drawing_a_triangle/02_Graphics_pipeline_basics/01_Shader_modules.adoc
+++ b/en/03_Drawing_a_triangle/02_Graphics_pipeline_basics/01_Shader_modules.adoc
@@ -160,12 +160,10 @@ float4 fragMain() : SV_Target
 
 The `fragMain` entry point function is called for every fragment just like the
 vertex shader `vertMain` function is called for every vertex. Colors in Slang
-are 4-component vectors with the R, G, B and alpha channels within the [0, 1] ranges. Unlike
-`SV_Position` in the vertex shader, there is no built-in variable to output a
-color for the current fragment. You have to specify your own output variable for
-each framebuffer where the `SV_TARGET` annotation specifies the index
-of the framebuffer. The color red is written to this `outColor` variable that is
-linked to the first (and only) framebuffer at index `0`.
+are 4-component vectors with the R, G, B and alpha channels within the [0, 1] range.
+The `SV_Target` annotation on the return value marks it as the color written to a
+framebuffer; since we only have a single framebuffer here, it implicitly targets
+index `0`. The color red is returned directly as `float4(1.0, 0.0, 0.0, 1.0)`.
 
 == Per-vertex colors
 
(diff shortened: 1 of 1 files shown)


SaschaWillems:
SV_Target is fully optional an Slang and we could as well just completely remove it. Slang will auto infer the render target, even if you use multiple ones.

PR #474, opened 2026-08-20
Check presentation support in isDeviceSuitable

### LINKED ISSUE #457: Should isDeviceSuitable() also check presentation support in the Window Surface chapter?
In [05_window_surface.cpp](https://docs.vulkan.org/tutorial/latest/_attachments/05_window_surface.cpp), `isDeviceSuitable()` checks whether the physical device has a queue family that supports graphics:

```c++
bool supportsGraphics =
    std::ranges::any_of(
        queueFamilies,
        [](auto const& qfp) {
            return !!(qfp.queueFlags &
                      vk::QueueFlagBits::eGraphics);
        });
```

However, createLogicalDevice() later requires a queue family that supports both graphics and presentation to the selected surface:

```c++
if ((queueFamilyProperties[qfpIndex].queueFlags &
     vk::QueueFlagBits::eGraphics) &&
    physicalDevice.getSurfaceSupportKHR(
        qfpIndex, *surface))
{
    queueIndex = qfpIndex;
    break;
}
```

It seems possible for a physical device to pass isDeviceSuitable() because it has a graphics-capable queue family, but then fail in createLogicalDevice() because no queue family supports both graphics and presentation to the surface.

Would it make sense for isDeviceSuitable() to also check presentation support, so that its suitability criteria match the requirements later used by createLogicalDevice()?

gpx1000:
isDeviceSuitable() only checked for a graphics-capable queue family, but createLogicalDevice() later requires one that supports both graphics and presentation. A device could pass suitability and then throw in createLogicalDevice(). Extends the existing search to also require getSurfaceSupportKHR(qfpIndex, *surface), applied consistently across every chapter from window-surface creation onward.

Fixes #457.

### DIFF
--- a/attachments/05_window_surface.cpp
+++ b/attachments/05_window_surface.cpp
@@ -166,9 +166,17 @@ class HelloTriangleApplication
 		// Check if the physicalDevice supports the Vulkan 1.3 API version
 		bool supportsVulkan1_3 = physicalDevice.getProperties().apiVersion >= VK_API_VERSION_1_3;
 
-		// Check if any of the queue families support graphics operations
+		// Check if any of the queue families support both graphics and presentation to our surface
 		auto queueFamilies    = physicalDevice.getQueueFamilyProperties();
-		bool supportsGraphics = std::ranges::any_of(queueFamilies, [](auto const &qfp) { return !!(qfp.queueFlags & vk::QueueFlagBits::eGraphics); });
+		bool supportsGraphicsAndPresent = false;
+		for (uint32_t qfpIndex = 0; qfpIndex < queueFamilies.size(); qfpIndex++)
+		{
+			if ((queueFamilies[qfpIndex].queueFlags & vk::QueueFlagBits::eGraphics) && physicalDevice.getSurfaceSupportKHR(qfpIndex, *surface))
+			{
+				supportsGraphicsAndPresent = true;
+				break;
+			}
+		}
 
 		// Check if all required physicalDevice extensions are available
 		auto availableDeviceExtensions = physicalDevice.enumerateDeviceExtensionProperties();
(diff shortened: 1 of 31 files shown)


attachments/05_window_surface.cpp
```diff
@@ -166,9 +166,17 @@ class HelloTriangleApplication
...
-		bool supportsGraphics = std::ranges::any_of(queueFamilies, [](auto const &qfp) { return !!(qfp.queueFlags & vk::QueueFlagBits::eGraphics); });
+		bool supportsGraphicsAndPresent = false;
+		for (uint32_t qfpIndex = 0; qfpIndex < queueFamilies.size(); qfpIndex++)
+		{
+			if ((queueFamilies[qfpIndex].queueFlags & vk::QueueFlagBits::eGraphics) && physicalDevice.getSurfaceSupportKHR(qfpIndex, *surface))
+			{
+				supportsGraphicsAndPresent = true;
+				break;
+			}
+		}
```
asuessenbach, line 179: Maybe a bit tricky, but what do you think about
```
bool supportsGraphicsAndPresent =
    std::ranges::any_of(queueFamilies,
                        [&physicalDevice, &surface = this->surface](auto const &qfp) {
	                        static int qfpIndex = 0;
	                        return physicalDevice.getSurfaceSupportKHR(qfpIndex++, *surface) && (qfp.queueFlags & vk::QueueFlagBits::eGraphics);
                        });
```

asuessenbach (approved):

attachments/05_window_surface.cpp
gpx1000: Not a bad idea.  I would normally worry about making it too hard for people to read that aren't adept with modern C++ idioms, but the Rubicon has already been crossed on that thought; I'll get started on updating.

gpx1000: Opting to use a function local counter captured by reference instead.  Same idea though.

SaschaWillems (approved):
Even though I don't like that syntax style at all, it works as expected. So I'm fine with it.

PR #478, opened 2026-08-21
Explain the IntelliSense/C++20-module preprocessor block in Base code

### LINKED ISSUE #141: information missmatch/missing after adding modules and raii
if you go to the first chapter's attachments here https://docs.vulkan.org/tutorial/latest/_attachments/00_base_code.cpp ...
you will notice the inclusion of these few lines of code:

```cpp
#ifdef __INTELLISENSE__
#include <vulkan/vulkan_raii.hpp>
#else
import vulkan_hpp;
#endif
```
which stop intelissense from throwing errors with the `import` token.

however, this is never actually mentioned on the tutorial at all, so viewers might stumble upon this and get stuck.

also, there are still these lines on chapter 0:
```cpp
#define GLFW_INCLUDE_VULKAN
#include <GLFW/glfw3.h>
```
despite now, we using the vulkan module, which (from my understanding, considering these lines arent on the attachment) invalidates the need of including vulkan in glfw3.

gpx1000:
The first code block readers see has an unexplained __INTELLISENSE__/USE_CPP20_MODULES guard. Adds a short explanation.

Fixes #141.

### DIFF
--- a/en/03_Drawing_a_triangle/00_Setup/00_Base_code.adoc
+++ b/en/03_Drawing_a_triangle/00_Setup/00_Base_code.adoc
@@ -41,13 +41,16 @@ class HelloTriangleApplication {
 ----
 
 The `#if defined(__INTELLISENSE__) || !defined(USE_CPP20_MODULES)` block picks
-between two ways of pulling in the Vulkan-Hpp API. When `USE_CPP20_MODULES` is
-defined, we `import vulkan_hpp;` as a C++20 module. Visual Studio and VS Code's
-IntelliSense engine doesn't yet understand `import` for modules and will flag
-it as an error even though the code compiles fine, so we fall back to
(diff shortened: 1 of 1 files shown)


en/03_Drawing_a_triangle/00_Setup/00_Base_code.adoc
```diff
@@ -38,6 +38,19 @@ import vulkan_hpp;
 #include <cstdlib>
 
 class HelloTriangleApplication {
+----
+
+The `#if defined(__INTELLISENSE__) || !defined(USE_CPP20_MODULES)` block picks
```
SaschaWillems, line 43: A note on modules and this define is already present a few lines below (92). So maybe consolidate both into one block or remove one of them.

SaschaWillems (changes requested):

SaschaWillems (approved):

PR #480, opened 2026-08-21
Fix parameter naming mismatch and add the filtered debugCallback example

### LINKED ISSUE #256: Lesson 3 Validation Layers Feedback
I have been going through this tutorial and have some feedback for it.

1. In the section "Using validation layers",  there is this code snippet:

```cpp
vk::InstanceCreateInfo createInfo{
    .pApplicationInfo        = &appInfo,
    .enabledLayerCount       = static_cast<uint32_t>(requiredLayers.size()),
    .ppEnabledLayerNames     = requiredLayers.data(),
    .enabledExtensionCount   = 0,
    .ppEnabledExtensionNames = nullptr };
```

The `enabledExtensionCount` and `ppEnabledExtensionNames` have been changed to 0 and nullptr, respectively, from the previous lesson.
This caused some confusion, so I would suggest the following:

```diff
vk::InstanceCreateInfo createInfo{
    .pApplicationInfo        = &appInfo,
    .enabledLayerCount       = static_cast<uint32_t>(requiredLayers.size()),
    .ppEnabledLayerNames     = requiredLayers.data(),
+  .enabledExtensionCount   =  glfwExtensionCount,
+  .ppEnabledExtensionNames = glfwExtensions };
```

And/or add an explanation that `enabledExtensionCount` and `ppEnabledExtensionNames`  will be updated later in the lesson.

2. The debugCallback() parameter is named `severity`, but in the snippet showing you can compare this enumeration, the parameter 
is named `messageSeverity`.  I suggest changing it to `severity` to match the function argument. 

```diff
+if (severity >= vk::DebugUtilsMessageSeverityFlagBitsEXT::eWarning) {
    // Message is important enough to show
}
````

3. The final code at the end, `debugCallback()` does a comparison on the severity parameter to only show warnings and error messages.
I suggest adding another snippet with this change with an explanation to make it more clear.

```cpp
static VKAPI_ATTR vk::Bool32 VKAPI_CALL debugCallback(vk::DebugUtilsMessageSeverityFlagBitsEXT severity, vk::DebugUtilsMessageTypeFlagsEXT type, const vk::DebugUtilsMessengerCallbackDataEXT *pCallbackData, void *)
	{
		if (severity == vk::DebugUtilsMessageSeverityFlagBitsEXT::eError || severity == vk::DebugUtilsMessageSeverityFlagBitsEXT::eWarning)
		{
			std::cerr << "validation layer: type " << to_string(type) << " msg: " << pCallbackData->pMessage << std::endl;
		}

		return vk::False;
	}
```

Also, is it possibly to just have the comparison be `severity >= vk::DebugUtilsMessageSeverityFlagBitsEXT::eWarning` like the snippet I show in 2.?

4.  Is it possible to add line highlighting in the code snippets for lines that have been added/changed?  I see that [asciidoc](https://docs.asciidoctor.org/asciidoc/latest/verbatim/highlight-lines/) has this capability.  It was hard for me to initially see what had changed in the last code snippet of `createInstance()` in this lesson and some others.
This would make it easier to follow, I believe.

gpx1000:
Renames messageSeverity to severity to match the actual parameter name, and adds the missing filtered-debugCallback snippet that matches the attachment code.

Fixes #256.

### DIFF
--- a/en/03_Drawing_a_triangle/00_Setup/02_Validation_layers.adoc
+++ b/en/03_Drawing_a_triangle/00_Setup/02_Validation_layers.adoc
@@ -285,8 +285,28 @@ some level of severity, for example:
 
 [,c++]
 ----
-if (messageSeverity >= vk::DebugUtilsMessageSeverityFlagBitsEXT::eWarning) {
+if (severity >= vk::DebugUtilsMessageSeverityFlagBitsEXT::eWarning) {
     // Message is important enough to show
+}
+----
+
+Applying that to our `debugCallback`, we can filter out anything less severe
+than a warning so that verbose and informational messages don't clutter the
+output:
+
+[,c++]
+----
+static VKAPI_ATTR vk::Bool32 VKAPI_CALL debugCallback(vk::DebugUtilsMessageSeverityFlagBitsEXT       severity,
+                                                      vk::DebugUtilsMessageTypeFlagsEXT              type,
+                                                      const vk::DebugUtilsMessengerCallbackDataEXT * pCallbackData,
+                                                      void *                                         pUserData)
+{
+  if (severity == vk::DebugUtilsMessageSeverityFlagBitsEXT::eWarning ||
+      severity == vk::DebugUtilsMessageSeverityFlagBitsEXT::eError) {
+    std::cerr << "validation layer: type " << to_string(type) << " msg: " << pCallbackData->pMessage << std::endl;
+  }
+
+  return vk::False;
 }
 ----
 
--- a/en/03_Drawing_a_triangle/00_Setup/02_Validation_layers.adoc
+++ b/en/03_Drawing_a_triangle/00_Setup/02_Validation_layers.adoc
@@ -310,6 +310,15 @@ static VKAPI_ATTR vk::Bool32 VKAPI_CALL debugCallback(vk::DebugUtilsMessageSever
 }
 ----
 
+Note that `severity` is comparable with `==` here rather than needing a
+bitwise `&`. It's typed as `vk::DebugUtilsMessageSeverityFlagBitsEXT`, not
+`vk::DebugUtilsMessageSeverityFlagsEXT`: the `FlagBits` suffix marks it as an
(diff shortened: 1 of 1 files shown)


en/03_Drawing_a_triangle/00_Setup/02_Validation_layers.adoc
```diff
@@ -285,11 +285,31 @@ some level of severity, for example:
...
+output:
+
+[,c++]
+----
+static VKAPI_ATTR vk::Bool32 VKAPI_CALL debugCallback(vk::DebugUtilsMessageSeverityFlagBitsEXT       severity,
+                                                      vk::DebugUtilsMessageTypeFlagsEXT              type,
+                                                      const vk::DebugUtilsMessengerCallbackDataEXT * pCallbackData,
+                                                      void *                                         pUserData)
+{
+  if (severity == vk::DebugUtilsMessageSeverityFlagBitsEXT::eWarning ||
```
SaschaWillems, line 304: Those are flags, so should be compared with "&" instead of "==".

SaschaWillems (changes requested):

en/03_Drawing_a_triangle/00_Setup/02_Validation_layers.adoc
asuessenbach: Nope.

Those are `FlagBits`, not `Flags`.
In vulkan terminology, a `FlagBits` is the enum, and each specified value is just one bit.
A `Flags`, on the other hand, is just an `uin32_t` and is used to hold zero or more corresponding `FlagBits`.

That is, using `==` instead of `&` here is correct.

SaschaWillems: Correct. Sorry for the mistake.

asuessenbach:
Maybe it's worth to explicitly mention the FlagBits vs. Flags handling here?

SaschaWillems (approved):

PR #504, opened 2026-09-05
Fixed small bug with RefCount

wpsimon09:
since `refCounts` member contains object with type `Resource` calling: `refCount[id]++` was incorrect. 

This PR fixes this issue.  However it also rises a question if we want to keep ref count within the `Resource` object or simply use `int` as a value of the `refCounts` member variable.

~~so basically there are two choices either we`ll have:~~

```c++
    std::unordered_map<std::type_index, std::unordered_map<std::string, ResourceData>>  refCounts;
```

or 

```c++
    std::unordered_map<std::type_index, std::unordered_map<std::string, int>>  refCounts;
```

~~The second approach would work with what was initially in the tutorial. Either way is fine so if you have time please let me know which version is better and I will alter this PR respectively.~~

Based on the rest of the tutorial it seems that niether of those types are correct and instead it is supposed to be 

```c++
std::unordered_map<std::string, ResourceData>  refCounts;
```

Thank you !

### DIFF
--- a/en/Building_a_Simple_Engine/Engine_Architecture/04_resource_management.adoc
+++ b/en/Building_a_Simple_Engine/Engine_Architecture/04_resource_management.adoc
@@ -134,6 +134,10 @@ private:
     struct ResourceData {
         std::shared_ptr<Resource> resource;  // The actual resource
         int refCount;                        // Reference count for this resource
+
+        // Utility functions 
+        void IncreaseRefCount() { refCount++; } 
+        void DecreaseRefCount() { refCount--; }
     };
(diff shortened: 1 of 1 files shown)


wpsimon09:
The rest of the code in the tutorial is structured with assumption that `refTypes` is 

```c++
    std::unordered_map<std::string, ResourceData>  refCounts;
``` 

I have altered the tutorial where it mattered to match this assumption. 

Furthermore some clarification was needed like for example why one needs `ResrouceHandle` if one can easily call `ResrouceManager::GetResrouce()`. This was fixed by using private specifier on those methods.

PR #515, opened 2026-09-16
Fix implicit unsigned to signed conversion

susgirl446:


asuessenbach (approved):

SaschaWillems (changes requested):
There are multiple occurrences of this in some of the advanced chapter C++ code files which should also be updated.

Issue #513, opened 2026-09-15
SIGGRAPH 2026 tutorial no longer compiles


SaschaWillems:
This was brought up in the Khronos forums with clangd, and I can reproduce it with Visual Studio 2026: When using a recent SDK (1.4.357), the " SIGGRAPH 2026: How to write a Vulkan application" code can't be compiled.

This seems to be an issue with a specific overload of the dynamic dispatcher in the Vulkan C++ headers.

Here is the full build error (sorry for the German text):

```
V:\Vulkan-Docs-Site\Vulkan-Tutorial-Main\attachments\siggraph2026_vk_tutorial\src\main.cpp(476,43): error C2661: "vk::detail::DispatchLoaderDynamic::init": Keine überladene Funktion akzeptiert 3 Argumente [V:\Vulkan-Docs-S
ite\Vulkan-Tutorial-Main\attachments\siggraph2026_vk_tutorial\build\vs2026\vulkan_siggraph.vcxproj]
      C:\VulkanSDK\1.4.357.0\include\vulkan\vulkan.hpp(27624,12):
      kann "void vk::detail::DispatchLoaderDynamic::init(const vk::Instance &,const vk::Device &) noexcept" sein
          V:\Vulkan-Docs-Site\Vulkan-Tutorial-Main\attachments\siggraph2026_vk_tutorial\src\main.cpp(476,43):
          „void vk::detail::DispatchLoaderDynamic::init(const vk::Instance &,const vk::Device &) noexcept“ erwartet, dass 2 Argumente – 3 angegeben werden.
      C:\VulkanSDK\1.4.357.0\include\vulkan\vulkan.hpp(27612,12):
      oder "void vk::detail::DispatchLoaderDynamic::init(const vk::Instance &,const vk::Device &,const DynamicLoader &) noexcept"
          V:\Vulkan-Docs-Site\Vulkan-Tutorial-Main\attachments\siggraph2026_vk_tutorial\src\main.cpp(476,79):
          "Initialisierung": "PFN_vkVoidFunction (__cdecl *)(VkInstance,const char *)" kann nicht in "const vk::Device &" konvertiert werden
              V:\Vulkan-Docs-Site\Vulkan-Tutorial-Main\attachments\siggraph2026_vk_tutorial\src\main.cpp(476,79):
              Ursache: Konvertierung von "overloaded-function" in "const vk::Device" nicht möglich
[...]

PR #517, opened 2026-09-17
Add CI for SIGGRAPH 2026 tutorial

SaschaWillems:
There is no CI yet for the SIGGRAPH 2026 tutorial, so if anything breaks (like in #513), we're not noticing that. This PR adds CI for the SIGGRAPH 2026 tutorial to close that gap.

Note: Currently a draft.


SaschaWillems:
Interestingly this fails on Linux, not due to an issue with CI, but an actual issue with the code itself. So adding CI for this already seems to show benefits :)

iagoCL:
Thank you very much for adding the CI workflow. I think it will be useful going forward.

I tried building the tutorial again on my Linux machine. It initially compiled successfully, but after updating Clang I was able to partially reproduce the issue.

I can fix the part I reproduced with the following change:

```patch
diff --git a/attachments/siggraph2026_vk_tutorial/src/main.cpp b/attachments/siggraph2026_vk_tutorial/src/main.cpp
index e44ed50..f0f0d94 100644
--- a/attachments/siggraph2026_vk_tutorial/src/main.cpp
+++ b/attachments/siggraph2026_vk_tutorial/src/main.cpp
@@ -538,7 +538,7 @@ void Application::initVulkanVKB()

         // Store format and extent of the swapchain images for later use.
         {
-            m_swapchainFormat = vk::Format{m_vkbData.m_swapchain.image_format};
+            m_swapchainFormat = static_cast<vk::Format>(m_vkbData.m_swapchain.image_format);
             m_swapchainExtent = vk::Extent2D{
                 .width = m_vkbData.m_swapchain.extent.width,
                 .height = m_vkbData.m_swapchain.extent.height,
```

I am not sure why this does not trigger on Windows, but I think we can simply fix it.

When using Vulkan, I do not think we should have to specify every initializer. Designated initializers make Vulkan code significantly less verbose and allow us to omit `pNext` and other unnecessary fields.

In theory, `-Wno-missing-designated-field-initializers` should solve the problem by allowing unused fields to be omitted when designated initializers are used.

Which version of Clang is installed in CI?

> `warning: unknown warning option '-Wno-missing-designated-field-initializers'; did you mean '-Wno-missing-field-initializers'? [-Wunknown-warning-option]`

As a workaround, we could also disable `-Wmissing-field-initializers` for the Linux Clang build, although I do not think this should be necessary:

```patch
diff --git a/attachments/siggraph2026_vk_tutorial/CMakeLists.txt b/attachments/siggraph2026_vk_tutorial/CMakeLists.txt
index 88cdda9..734b66f 100644
--- a/attachments/siggraph2026_vk_tutorial/CMakeLists.txt
+++ b/attachments/siggraph2026_vk_tutorial/CMakeLists.txt
@@ -293,7 +293,7 @@ if(MSVC)
 else()
   target_compile_options(vulkan_siggraph PRIVATE -Wall -Wextra -Wpedantic)
   if(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
-    target_compile_options(vulkan_siggraph PRIVATE -Wno-missing-designated-field-initializers)
+    target_compile_options(vulkan_siggraph PRIVATE -Wno-missing-designated-field-initializers -Wno-missing-field-initializers)
   endif()
 endif()
```

SaschaWillems:
Looking at the CI failure, I don't think that the issue is caused by the missing initializers. That does cause a few warnings, but the error is an actual type conversion error:

```/home/runner/work/Vulkan-Tutorial/Vulkan-Tutorial/attachments/siggraph2026_vk_tutorial/src/main.cpp:541:44: error: non-constant-expression cannot be narrowed from type 'VkFormat' to 'vk::Format' in initializer list [-Wc++11-narrowing]
  541 |             m_swapchainFormat = vk::Format{m_vkbData.m_swapchain.image_format};
```

iagoCL:
Yes, you are right. The CI failure is caused by a type-conversion error. For some reason, it does not occur on Windows. The first change I suggested fixes it.

I have added the fix in a new branch based on this PR:
- https://github.com/iagoCL/Vulkan-Tutorial/tree/iagcal01/siggraph_tutorial_linux_compilation
- Working version: https://github.com/iagoCL/Vulkan-Tutorial/commit/af8db2cd9fa4950f2c90ef67a9770398982b2523
- I also adjust the CI steps as we do not need the same libraries https://github.com/iagoCL/Vulkan-Tutorial/commit/041998225a8dd265c181e3b6b6e0656e8b3c17da

This change fixes the Linux build:

```diff
-            m_swapchainFormat = vk::Format{m_vkbData.m_swapchain.image_format};
+            m_swapchainFormat = static_cast<vk::Format>(m_vkbData.m_swapchain.image_format);
```

I also added compiler-version-specific flags to suppress the relevant Clang warnings and documented the build dependencies in the README.
I verified the changes in CI and tested them with Clang 17 through Clang 21 on both Linux and Windows.

SaschaWillems:
Thank you very much 👍🏻

CI is now passing, so this is ready to review :)

gpx1000 (approved):

